PR #11, opened 2015-06-16
Improve winapi usage

retep998:
Switch to relying on `kernel32` for `ReOpenFile`.

As well `winapi` and `kernel32` are no-ops on non-windows platforms so just depend on them unconditionally.


Stebalien:
Thanks!

PR #15, opened 2015-11-03
introduce new 2 functions

KeenS:
I'd like to have 2 new functions `template` and `template_in` at NamedTempFile. Having some controls of file name brings flexibility.


KeenS:
It seems CI is failing on windows with newer rustc but I don't touch windows specific code so much. How can I fix build?.l

Stebalien:
Unrelated. Try rebasing.

KeenS:
Done. CI all passing. Thank you.

Stebalien:
I'd rather not do any template parsing at runtime. 

Maybe use the builder pattern to create the template?

```rust
CustomNamedTempfile::new().prefix("tmp").postfix("abcde").rand(5).new_in(...);
```

Alternatively, provide a macro and abuse format strings (requires rust 1.5+):

```rust
macro_rules! tempfile {
    ($fmt:tt) => {{
        struct Formatter;
        use ::std::fmt;
        impl fmt::Display for Formatter {
            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                let width = f.width().unwrap_or(6);
                // ...Write random characters...
                Ok(())
            }
        }
        // templated should probably be hidden.
        TempFile::templated(format_args!($fmt, Formatter))
    }}
}

fn main() {
    let file = tempfile!("/path/to/tmp{:6}.log"); // no leading '/' means put it in /tmp.
}
```

KeenS:
That's pretty way. I was C minded.

KeenS:
@Stebalien 
Let me have a talk with you.
I implemented a builder pattern but left a TODO.
Because I'm not so familiar with Rust, I don't know the Rust way to treat errors in
method chains. How should it be done? Wrapping every chained methods with `map` will be terrible.

And one more thing is do you know how can I hide `tmpname` from users without exposing any fields of CustomNamedTempFile? One option may be providing public method `path` and converting returned `Path` to `OsString` if needed. However, since return value can vary on each method call because of randomness, it is very misleading (importantly, the returned value will deffer from created temp file's name.).

I'll left to you to provide higher interface such as proposed `tempfile` macro. I'm too new to Rust to design useful APIs

src/named.rs
```diff
@@ -211,3 +205,108 @@ impl std::os::windows::io::AsRawHandle for NamedTempFile {
...
+pub struct CustomNamedTempFile {
+    random_len: usize,
+    prefix: String,
+    postfix: String    
+}
+
+impl CustomNamedTempFile {
+
+    /// Start building CustomNamedTempFile. See `new` for more information.
+    pub fn start() -> Self {
```
Stebalien: Usually called `new`.

src/named.rs
```diff
@@ -211,3 +205,108 @@ impl std::os::windows::io::AsRawHandle for NamedTempFile {
...
+        CustomNamedTempFile {
+            random_len: ::NUM_RAND_CHARS,
+            prefix: ".".to_string(),
```
Stebalien: I'd rather not allocate by default. Maybe use Cow?

src/named.rs
KeenS: Yes, I know but it conflicts the name of the function creating a tempfile. AFAK Rust doesn't have overloading. I caught up some options

1. `new` to allocate the builder and `create` to create a tempfile
2. `builder` to allocate the builder and `new` to create a tempfile

Do you think the 1st option is the best way?

src/named.rs
KeenS: Actually it doesn't require ownership, I decided to use `&str`

Stebalien:
Sorry for the delay, I'll get on this ASAP. I'm a bit busy with thesis/TAing at the moment.

Stebalien:
Merged with some API changes. Thanks!

KeenS:
:sushi:

PR #22, opened 2016-01-18
imp/unix: always cast cstr.as_ptr() to avoid issues with differing c_char types

jmesmon:
Ran into this issue on arm. Unclear why they differ, but I'd imagine it's due to a mismatch between rustc's libc and the libc crate.


Stebalien:
This should be fixed in a future version of stdlib...

PR #26, opened 2016-04-21
Expose PersistError

### LINKED ISSUE #25: PersistError is private
This seems to make it hard to match on errors from `persist`, or perhaps my Rust is just not strong enough?

I want to write

```rust
    if let tempfile::PersistError(e) = f.persist_noclobber(path) {
        return Err(e.error);
    };
```

sourcefrog:
Fixes #25

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,5 +42,5 @@ mod imp;
 mod named;
 mod unnamed;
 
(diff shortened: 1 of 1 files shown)


Stebalien:
Oops. Thanks.

PR #32, opened 2017-08-12
Add Redox OS backend

ids1024:


Stebalien:
Nice! Thanks.

Unfortunately, that's a *lot* of duplicate code. Given that Redox is unix-like(ish), do you think you can merge this into the unix module (maybe introducing some form of general-purpose `cvt_err` function to abstract away the differences in the syscall library)?

Stebalien:
Alternatively, we could also add a `nixc` submodule that abstracts over the various low-level unix libc functions we need.

ids1024:
In addition to errors, the Redox functions differ in that they accept u8 slices rather than pointers to c strings. I guess having functions wrapping those two things would probably make it possible to merge this with the unix implementation...

Stebalien:
My primary concern is that I can't test Redox easily so any non-trivial Redox only logic will likely experience severe bitrot.

ids1024:
Okay, I've merged the Redox code into `imp/unix.rs`.

src/imp/unix.rs
```diff
@@ -99,24 +114,30 @@ pub fn reopen(file: &File, path: &Path) -> io::Result<File> {
...
             let _ = unlink(old_path.as_ptr() as *const c_char);
         }
         Ok(())
     }
 }
+
+#[cfg(target_os = "redox")]
+pub fn persist(old_path: &Path, new_path: &Path, overwrite: bool) -> io::Result<()> {
+    // XXX implement in better way when possible
+    if !overwrite && new_path.exists() {
```
Stebalien: Is there no way to do this atomically (avoid clobbering existing files)? We're currently guaranteeing that this is atomic.

src/imp/unix.rs
```diff
@@ -28,15 +38,23 @@ pub fn cstr(path: &Path) -> io::Result<CString> {
...
+                                   0o600)));
+        Ok(FromRawFd::from_raw_fd(fd))
+    }
+}
+
+#[cfg(target_os = "redox")]
+pub fn create_named(path: &Path) -> io::Result<File> {
+    unsafe {
+        let fd = try!(cvt_err(open(path.as_os_str().as_bytes(),
+                                   O_CLOEXEC | O_EXCL | O_RDWR | O_CREAT | 0o600)));
```
Stebalien: I'm pretty sure the `0o600` is not supposed to be ORed with this. Is there any way to atomically set the permissions on this file?

src/imp/unix.rs
ids1024: I don't think so; this can be fixed once that is possible.

It's actually somewhat worse; `rename` is not atomic either currently (just copy and unlink); that really should be implemented, but it hasn't been done yet.

src/imp/unix.rs
ids1024: This should do that on Redox: https://github.com/redox-os/redoxfs/blob/master/mount/redox/scheme.rs#L282

Ping @jackpot51

src/imp/unix.rs
Stebalien: I thought I saw a link syscall... my primary concern is not clobbering existing files and I'd rather just always return a not supported error to the user than potentially overwrite a file.

ids1024: There is a `link` function in redox_syscall, but hard links are not actually implemented yet.

Stebalien: Wait... by "copy and unlink" you mean it literally *copies* the file byte-by-byte? If that's the case, I think the only sane implementation here would be to return an error. The entire point of this method is that it allows the programmer to atomically replace an existing file with an updated file (and I've seen users rely on this guarantee).

src/imp/unix.rs
jackpot51: Yes, `open` on Redox includes mode flags in the flags parameter.

Stebalien: Got it. Thanks!

src/imp/unix.rs
```diff
@@ -8,18 +9,27 @@ use std::path::Path;
...
-#[cfg(not(all(lfs_support, target_os = "linux")))]
-use libc::stat as stat_t;
+#[cfg(not(any(all(lfs_support, target_os = "linux"), target_os = "redox")))]
+use libc::{open, fstat, stat as stat_t};
+
+#[cfg(target_os = "redox")]
+use syscall::{self, open, fstat, Stat as stat_t, O_EXCL, O_RDWR, O_CREAT, O_CLOEXEC};
+
+#[cfg(not(target_os = "redox"))]
+pub fn cvt_err(result: c_int) -> io::Result<c_int> {
```
Stebalien: Can we mark this as `#[inline(always)]`?

src/imp/unix.rs
```diff
@@ -8,18 +9,27 @@ use std::path::Path;
...
+    if result == -1 {
+        Err(io::Error::last_os_error())
+    } else {
+        Ok(result)
+    }
+}
+
+#[cfg(target_os = "redox")]
+pub fn cvt_err(result: Result<usize, syscall::Error>) -> io::Result<usize> {
```
Stebalien: Same.

ids1024:
Fair enough; I've made persist return an error on Redox; that can be implemented in the same way as Unix once Redox supports `rename` and `link`.

I also added `#[inline(always)]`.

Stebalien:
Thanks!

ids1024:
@Stebalien Would you mind doing a release, so Redox can use this without a Cargo.toml patch for the crate using this library?

Stebalien:
@ids1024 done. Sorry, it took so long; I'm traveling and don't have the best internet at the moment.

PR #33, opened 2017-09-12
Merge tempdir api

KodrAus:
Combines the [`tempdir`](https://github.com/rust-lang-nursery/tempdir) API, beefs out the docs, and refactors `NamedTempFile` a bit.

If you'd rather I didn't refactor the internal structure then I can rejig this PR to minimise file movement.

Some previous discussion/iteration on combining the APIs: https://github.com/KodrAus/tempfile/pull/1

cc: @Stebalien @alexcrichton


KodrAus:
@Stebalien you'll probably want to take a thorough look through the docs I've added, I might've gotten some security notices wrong.

KodrAus:
Support for `1.13.0` needs a [quick change in `remove_dir_all`](https://github.com/Aaronepower/remove_dir_all/pull/2)

Stebalien:
I will review this ASAP but that probably won't be till the weekend.

KodrAus:
Friendly ping @Stebalien to take a look at this when you've got bandwidth. No pressure though.

We could move forward without that fix in `remove_dir_all` but it would mean bumping the minimum rust version supported.

opilar:
A little question: when `tempdir` and `tempfile` will merge apis, should we rename the crate to `temp`?

KodrAus:
@opilar the idea is to keep the `tempfile` crate name.

Stebalien:
@KodrAus sorry, my family ended up visiting that weekend and now I'm traveling but I'll try to get to this in the next few days.

Stebalien:
I'm still alive, FYI (just busy). Sorry.

KodrAus:
@stebalien that's no problem :) We're not working on a tight timeline here so there's no hurry, no pressure. I just remembered there was a new version of `remove_dir_all` I could use.

azriel91:
ping! I came across this when looking to write tests that use a temporary directory. Is this PR close to being merged, or shall I use `tempdir`?

KodrAus:
@azriel91 sorry I just got a chance to loop back here. I would use `tempdir`. We'll have a nice smooth upgrade path when the APIs are merged, but this PR is only one step of that process.

src/file/mod.rs
```diff
@@ -0,0 +1,807 @@
...
+    pub fn as_file(&self) -> &File {
+        &self.inner().file
+    }
+
+    /// Get a mutable reference to the underlying file.
+    pub fn as_file_mut(&mut self) -> &mut File {
+        &mut self.inner_mut().file
+    }
+
+    /// Convert the temporary file into a `std::fs::File`.
```
Stebalien: This *should* note that the inner file will be deleted.

KodrAus:
Thanks for the thorough review @Stebalien! I'll rebase and work through the changes over the next week :+1:

KodrAus:
I've pushed up some changes based on the feedback because having code only living on my local machine makes me nervous...

I haven't updated the doc tests yet, and the behaviour of `TempDir` is now subtly different with respect to leading `.`s in the directory name. I think it would be good to think a bit about what we might want to do with the `Builder` in the future, and whether we're painting ourselves into any corners by sharing it between temporary files and directories.

Stebalien:
I'll try to take a look at it this weekend (and not just put this off for a few months this time...).

Note: I'm not entierly against changing the default for tempfile if it's a problem for tempdir. IIRC, we never made any guarantees (unless the user explicitly specified the prefix).

KodrAus:
No problem! I'll try get the tests and docs updated before then.

src/dir.rs
```diff
@@ -0,0 +1,272 @@
...
+/// directory are dropped before the `TempDir` goes out of scope. The
+/// `TempDir` destructor will silently ignore any errors in deleting
+/// the directory; to instead handle errors call [`TempDir::close`].
+///
+/// Note that if the program exits before the `TempDir` destructor is
+/// run, such as via [`std::process::exit`], by segfaulting, or by
+/// receiving a signal like `SIGINT`, then the temporary directory
+/// will not be deleted.
+/// 
+// # Examples
```
jasonwhite: Missing a triple slash (`///`) here.

src/dir.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// drop(tmp_file);
+    /// tmp_dir.close()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn close(mut self) -> io::Result<()> {
+        let result = remove_dir_all(self.path());
+
+        // Prevent the Drop impl from removing the dir a second time.
+        self.path = None;
```
jasonwhite: If the entire reason for wrapping the path in an `Option` is to prevent `Drop` from attempting to delete the directory a second time, then you should be able to eliminate the need for `Option` with this:

```rust
pub fn close(mut self) -> io::Result<()> {
    // Replace the path with an empty one to avoid a memory leak by mem::forget.
    let path = std::mem::replace(self.path, PathBuf::new());
    std::mem::forget(self); // Don't call drop
    remove_dir_all(&path)
}
```

The same would have to be done to `into_path()` of course.

This is cleaner than using an `Option` in my opinion.

Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "tempfile"
-version = "2.2.0"
-authors = ["Steven Allen <steven@stebalien.com>"]
-description = "Securely create temporary files."
+version = "2.1.6"
```
jasonwhite: Looks like a package version number regression here.

src/lib.rs
```diff
@@ -1,36 +1,95 @@
...
+//! ```
+//!
+//! [`tempfile`]: fn.tempfile.html
+//! [struct.TempDir]: struct.TempDir.html
+//! [`NamedTempFile`]: struct.NamedTempFile.html
+//! [`std::env::temp_dir()`]: https://doc.rust-lang.org/std/env/fn.temp_dir.html
 
+#![doc(html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
+       html_favicon_url = "https://www.rust-lang.org/favicon.ico",
+       html_root_url = "https://docs.rs/tempfile/2.1.6")]
```
jasonwhite: This version number should match the one in `Cargo.toml` (once the one in there is fixed).

src/dir.rs
jasonwhite: Another problem to consider is what to do if `close()` fails. The caller might wish to implement a retry if, for example, they've created a temporary directory on a network file system that might have intermittent failures. (I encounter this problem a lot with network shares under Windows.)

However, it looks like you want API compatibility with the `tempdir` crate, so you probably can't solve this problem.

Cargo.toml
```diff
@@ -12,6 +14,7 @@ license = "MIT/Apache-2.0"
 
 [dependencies]
 rand = "0.4"
+remove_dir_all = "0.3"
```
jasonwhite: Looks like you can use `remove_dir_all` from stable Rust now. https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html

src/dir.rs
Stebalien: > The caller might wish to implement a retry if, for example, they've created a temporary directory on a network file system that might have intermittent failures. 

...

>  It looks like this problem has already been talked to death in

Also, in posix, one should generally *not* retry a close. The correct way to handle such errors is to catch them with an fsync first.

jasonwhite: @Stebalien Are you talking about `close(2)` or `remove_dir_all`? I agree about `close(2)`, but deleting files over a network can have intermittent failures where retrying is the right thing to do.

Stebalien: Ah. I was thinking about close. You're right, this is a bit different. However, I kind of feel that this is, really, the best API. We could put the path into the error but that'll make this less ergonomic.

Cargo.toml
KodrAus: Ah well spotted!

src/dir.rs
```diff
@@ -157,68 +95,8 @@ impl TempDir {
...
-            match fs::create_dir(&path) {
-                Ok(_) => return Ok(TempDir { path: Some(path) }),
-                Err(ref e) if e.kind() == ErrorKind::AlreadyExists => {}
-                Err(e) => return Err(e),
-            }
-        }
-
-        Err(Error::new(ErrorKind::AlreadyExists,
-                       "too many temporary directories already exist"))
+    pub fn new() -> io::Result<TempDir> {
```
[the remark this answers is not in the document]
KodrAus: Ah yes, I did misunderstand, I'll restore those methods :+1:

src/dir.rs
```diff
@@ -355,3 +233,8 @@ impl Drop for TempDir {
         }
     }
 }
+
+// pub(crate)
+pub fn create(path: PathBuf) -> io::Result<TempDir> {
```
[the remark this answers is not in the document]
KodrAus: Yep, that was another misunderstanding, I'll restore these too.

src/dir.rs
```diff
@@ -0,0 +1,272 @@
...
+impl Drop for TempDir {
+    fn drop(&mut self) {
+        // Path is `None` if `close()` or `into_path()` has been called.
+        if let Some(ref p) = self.path {
+            let _ =  remove_dir_all(p);
+        }
+    }
+}
+
+// pub(crate)
```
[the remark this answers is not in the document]
KodrAus: That's right, it's just a note that this method is accessible throughout the crate, but not public since our min Rust version doesn't support `pub(restricted)`.

src/lib.rs
```diff
@@ -108,8 +108,316 @@ extern crate syscall;
...
+    /// # fn run() -> Result<(), io::Error> {
+    /// # use tempfile::Builder;
+    /// let named_temp_file = Builder::new()
+    ///     .create()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// [`NamedTempFile::new`]: struct.NamedTempFile.html#method.new
+    pub fn named_tempfile(&self) -> io::Result<NamedTempFile> {
```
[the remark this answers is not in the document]
KodrAus: That's a fair point :+1:

KodrAus:
Thanks @jasonwhite and @Stebalien! I'll work through this round of feedback over this week.

Stebalien:
Awesome! Thanks for putting this together.

KodrAus:
@Stebalien @jasonwhite this should be ready for another look now.

Stebalien:
:tada:

@KodrAus thanks so much for both implementing this and then bearing with me on this long and drawn out CR.

@jasonwhite thanks for pitching in and waiting for this CR to complete.

KodrAus:
Thanks all!

I'll get the ball rolling on deprecating the `tempdir` crate. Looks like we've got an [open design question](https://github.com/rust-lang-nursery/tempdir/issues/44) in `tempdir` we might want to look at here but otherwise everything looks good on that end.

PR #34, opened 2017-12-31
bump rand to 0.4

ignatenkobrain:


KodrAus:
In #33 I've ended up [bumping the min version to `1.13`](https://github.com/Stebalien/tempfile/pull/33/files#diff-354f30a63fb0907d4ad57269548329e3R8). Does `rand 0.4` build on `1.13` too?

PR #37, opened 2018-01-10
Fix warning about unused variable

jasonwhite:


Stebalien:
That's fixed in #33 and I'd like to avoid merging anything else right now to avoid introducing *too* many conflicts (it's a massive PR). However, I'll leave this open until that gets merged (hopefully soon).

jasonwhite:
Okay. No problem.

PR #38, opened 2018-01-11
Allow closing of a NamedTempFile file handle

### LINKED ISSUE #36: Close NamedTempFile without deleting
Is there a way to close the file handle without deleting the actual file? On Windows, I have a problem where a child process cannot open the temporary file for reading because the file handle is open by the parent process, resulting in a sharing violation.
 
The child process (which I cannot change) is opening the file with *only* `FILE_SHARE_READ`, not `FILE_SHARE_READ | FILE_SHARE_WRITE`. According to the [Microsoft docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx), if `FILE_SHARE_WRITE` is not specified in the share mode and another process has the file open for writing, `CreateFile` will fail with a sharing violation. Thus, the only way to fix this is to close the file handle before spawning the child process.

jasonwhite:
Fixes #36.

I've tested this with my project that is using tempfile and it works as expected.

I will refactor after #33 is merged.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -1,9 +1,12 @@
 use std::io::{self, Read, Write, Seek, SeekFrom};
 use std::fs::{self, File};
 use std::path::{Path, PathBuf};
+use std::ops::Deref;
+use std::ffi::OsStr;
 use std::error;
 use std::fmt;
 use std::env;
+use std::mem;
 use std;
 
 use ::Builder;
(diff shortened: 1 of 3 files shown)


src/named.rs
```diff
@@ -391,3 +402,89 @@ impl<'a, 'b> NamedTempFileOptions<'a, 'b> {
...
+    ///
+    /// If a file exists at the target path, persist will atomically replace it.
+    /// Upon failure, `self` is returned in the error result.
+    ///
+    /// Note: Temporary files cannot be persisted across filesystems.
+    ///
+    /// *SECURITY WARNING:* Only use this method if you're positive that a
+    /// temporary file cleaner won't have deleted your file. Otherwise, you
+    /// might end up persisting an attacker controlled file.
+    pub fn persist<P: AsRef<Path>>(self, new_path: P) -> Result<(), (TempPath, io::Error)> {
```
jasonwhite: This error type is the only thing I'm not 100% happy about. I can't use the existing `PersistError` because it isn't general enough. Any ideas? Or is this fine?

src/named.rs
```diff
@@ -391,3 +402,89 @@ impl<'a, 'b> NamedTempFileOptions<'a, 'b> {
...
+        match imp::persist(&self.0, new_path.as_ref(), true) {
+            Ok(_) => Ok(()),
```
Stebalien: You'll need to keep the drop glue from running here (otherwise, we'll try to delete the file we've just moved).

src/named.rs
Stebalien: An alternative is to just provide a method for reopening the file (we can then persist from `NamedTempFile`). However, this would be slightly less efficient. Regardless, I'd like such a method anyways (e.g., `open()`).

We could also just create a new error type. I'll leave that up to you. However, I *wouldn't* just use a tuple like this; `PersistError` has some convenience conversion functions that are kind of useful.

src/named.rs
```diff
@@ -391,3 +402,89 @@ impl<'a, 'b> NamedTempFileOptions<'a, 'b> {
...
+    pub fn persist_noclobber<P: AsRef<Path>>(self, new_path: P) -> Result<(), (TempPath, io::Error)> {
+        match imp::persist(&self.0, new_path.as_ref(), false) {
+            Ok(_) => Ok(()),
+            Err(err) => Err((self, err)),
+        }
+    }
+}
+
+impl fmt::Debug for TempPath {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
Stebalien: Does `#[derive(Debug)]` not work? If not, I'd use the debug builder at least.

src/named.rs
jasonwhite: Great catch. `std::mem::forget` should fix that.

jasonwhite:
@Stebalien I ended up refactoring this quite a bit and cleaned up how `NamedTempFile` works internally. (Sorry if this makes it harder to review!)

It occurred to me that it would be a lot cleaner if `NamedTempFile` were simply composed of a `TempPath` and `File`. This shouldn't change the public API at all. It only simplifies the internal workings of `NamedTempFile`.

I haven't implemented `TempPath::open()`, but I will if you want me to. The implementation isn't as simple as it might first seem.

Stebalien:
*Very* nice job! Looks ready to merge once #33 is done. Thanks!

Stebalien:
@jasonwhite this is ready for a rebase and merge now, when you have a chance.

jasonwhite:
All done. 😉

Stebalien:
@jasonwhite awesome! Thanks!

PR #46, opened 2018-02-15
Bump remove_dir_all dependency

Ryman:
This removes winapi 0.2 from the dependency graph!


Ryman:
Ah, didn't see #45  !

PR #47, opened 2018-03-15
fix typo in src/file/mod.rs

bbigras:


Stebalien:
Ah. That's not actually a typo. "Iff" means "if and only if" (sufficient and necessary, not just sufficient). However, in this case, "unless there's already a file at the target path" would probably be clearer.

bbigras:
Sorry. There's so many typos on github projects that I assumed it was one.

Stebalien:
I'm *shocked*! I for, for one, *never* make typos. :smile:

---

(regardless, thanks for checking)

PR #52, opened 2018-05-10
assert that tempfile*() meets its guarantees

oliverhenshaw:
tmpfiles on unix systems without O_TMPFILE are not cleaned up after use. Fix this and add some tests to assert that tmpfile*() meets its guarantees with regard to security and resource leaking.

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -1,7 +1,7 @@
 #[cfg(not(target_os = "redox"))]
 use libc::{c_char, c_int, link, rename, unlink, O_CLOEXEC, O_CREAT, O_EXCL, O_RDWR};
 use std::ffi::CString;
-use std::fs::{File, OpenOptions};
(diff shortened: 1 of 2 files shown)


oliverhenshaw:
This was tested on linux, and the general unix behaviour was tested:
(a) by setting TMPDIR=/path/to/some/fat/filesystem/temp/ - as O_TMPFILE isn't supported on fat. This was quite slow;
(b) by commenting out the linux-specific src/file/imp/unix.rs:create() variant.

It hasn't been tested on a real non-linux unix or on windows.

The branch at https://github.com/oliverhenshaw/tempfile/tree/unlink-unix-testable has broken out patches for easier testing.

Stebalien:
(and thanks for the test)!

Stebalien:
I went ahead and yanked 3.0.0 and 3.0.1 (and published 3.0.2) as this is a kind of horrible bug for a temporary file library.

PR #55, opened 2018-05-17
fix up some errors in docs

### LINKED ISSUE #53: Undefined variables in documentation for TempDir::new_in
The docs for `TempDir::new_in` refer to `tmpdir` and `prefix`, neither of which is an argument to the function.  `tmpdir` should probably be `dir`, but I'm not sure about `prefix`.  The mention of `prefix` should probably just be removed.
https://docs.rs/tempfile/3.0.2/tempfile/struct.TempDir.html#method.new_in

KodrAus:
Fixes #53 

We had some lingering references in the prose to some old method arguments on `TempDir`.


Stebalien:
Thanks.

PR #61, opened 2018-06-22
Add parameter for not restricting access to the named tempfile

### LINKED ISSUE #30: Restore default permissions on persist (on Unix)
Like #2 and #3, it would be nice to get a normal file after calling `persist` on a `NamedTemporaryFile`. `NamedTemporaryFile` is opened with `0o600`, I'd expect to get a `0o644` file after that.

tbu-:
Fixes #30.

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -401,6 +401,6 @@ impl Drop for TempDir {
 }
 
 // pub(crate)
-pub fn create(world_accessible: bool, path: PathBuf) -> io::Result<TempDir> {
+pub fn create(_world_accessible: bool, path: PathBuf) -> io::Result<TempDir> {
     fs::create_dir(&path).map(|_| TempDir { path: Some(path) })
 }
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -401,6 +401,6 @@ impl Drop for TempDir {
 }
 
 // pub(crate)
-pub fn create(path: PathBuf) -> io::Result<TempDir> {
+pub fn create(world_accessible: bool, path: PathBuf) -> io::Result<TempDir> {
     fs::create_dir(&path).map(|_| TempDir { path: Some(path) })
 }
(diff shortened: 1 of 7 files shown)


tbu-:
This is just a PoC right now, will fix the warnings etc. if this can get merged in one way or another.

What do you think about this approach? I couldn't find a way to reset file permissions to default on unixy systems after having `open`ed the file.

This would fix my problem with #30, as I'd like to persist the temporary files to normal files after being done writing to them.

Stebalien:
Sounds reasonable. However, I'd rather accept an `std::fs::Permissions` object as that's more general and cross platform. Would that work?

tbu-:
Hm. There doesn't seem to be a way to open a file with a given `Permissions` struct. The struct is opaque except for the readonly bit which isn't helpful in this case (and the mode on unixy systems). This means we could currently only support setting the mode on new files on unixy systems (not even Redox, and certainly not on Windows). I don't think taking a `Permissions` struct would improve this API.

Stebalien:
Gah. I keep not posting my replies.

You're absolutely correct, that's unfortunate.

Stebalien:
So, thinking about this, I'm fine with this approach. I'd rather have a more general approach but... the only API this touches is the builder. Given that, I'm fine going ahead with this.

tbu-:
Fixed warnings, added some documentation, added a test.

tbu-:
Yay, now even compiles on Windows. It's good to go from my side now. :)

Stebalien:
This is still going to need Windows and, preferably, directory support.

quark-zju:
Any progress on this? I want this feature too. @tbu- Do you mind if I work on a new pull request?

tbu-:
@quark-zju Feel free to redo this. :)

lucab:
Cross-referencing from https://github.com/Stebalien/tempfile/issues/114: on Linux though it would be helpful to be able to pass a proper `Permissions` argument.

Stebalien:
Fixed in #273.

PR #63, opened 2018-07-01
Reduce scope of unsafe code

alexmaco:
There were a few low-hanging fruit with respect to unsafeness.

In one case (`reopen` on unix) unsafe is now completely unnecessary thanks to std.

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -106,17 +106,15 @@ unsafe fn stat(fd: RawFd) -> io::Result<stat_t> {
 
 pub fn reopen(file: &File, path: &Path) -> io::Result<File> {
     let new_file = OpenOptions::new().read(true).write(true).open(path)?;
-    unsafe {
-        let old_meta = stat(file.as_raw_fd())?;
-        let new_meta = stat(new_file.as_raw_fd())?;
-        if old_meta.st_dev != new_meta.st_dev || old_meta.st_ino != new_meta.st_ino {
-            return Err(io::Error::new(
-                io::ErrorKind::NotFound,
-                "original tempfile has been replaced",
-            ));
-        }
-        Ok(new_file)
+    let old_meta = unsafe { stat(file.as_raw_fd())? };
+    let new_meta = unsafe { stat(new_file.as_raw_fd())? };
+    if old_meta.st_dev != new_meta.st_dev || old_meta.st_ino != new_meta.st_ino {
+        return Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "original tempfile has been replaced",
+        ));
     }
+    Ok(new_file)
 }
 
 #[cfg(not(target_os = "redox"))]
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -98,21 +98,21 @@ fn create_unix(dir: &Path) -> io::Result<File> {
     })
 }
 
-#[cfg(not(unix))]
+#[cfg(target_os = "redox")]
 unsafe fn stat(fd: RawFd) -> io::Result<stat_t> {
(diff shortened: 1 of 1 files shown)


src/file/imp/unix.rs
```diff
@@ -98,32 +98,44 @@ fn create_unix(dir: &Path) -> io::Result<File> {
     })
 }
 
+#[cfg(not(unix))]
```
Stebalien: This is a unix only module so this shouldn't be necessary.

src/file/imp/windows.rs
```diff
@@ -92,22 +92,22 @@ pub fn persist(old_path: &Path, new_path: &Path, overwrite: bool) -> io::Result<
...
-        let old_path_w = to_utf16(old_path);
-        let new_path_w = to_utf16(new_path);
+    let old_path_w = to_utf16(old_path);
+    let new_path_w = to_utf16(new_path);
+
+    let flags = if overwrite {
+        MOVEFILE_REPLACE_EXISTING
+    } else {
+        0
+    };
```
Stebalien: While this function is reasonably short, moving the flags away from the call site makes it a harder to read (less immediately obvious where the flags apply).

IMO, readability is significantly more important than reducing the amount of code in an unsafe block.

src/file/imp/windows.rs
```diff
@@ -92,22 +92,22 @@ pub fn persist(old_path: &Path, new_path: &Path, overwrite: bool) -> io::Result<
     // TODO: We should probably do this in one-shot using SetFileInformationByHandle but the API is
     // really painful.
 
-    unsafe {
-        let old_path_w = to_utf16(old_path);
-        let new_path_w = to_utf16(new_path);
+    let old_path_w = to_utf16(old_path);
+    let new_path_w = to_utf16(new_path);
```
Stebalien: As the types here do not ensure that these vectors are safe to pass into the kernel as c strings, this should go in the unsafe block.

src/file/imp/unix.rs
alexmaco: It looks like this module is also used for redox. Does `target_os == "redox"`` always imply `unix` ?

Stebalien: Ah. I forgot about that... You're right. Can we say "cfg(redox)" and "cfg(not(redox))"?

alexmaco: Yes, that would be clearer than `unix` and `not(unix)`.

src/file/imp/unix.rs
```diff
@@ -4,15 +4,15 @@ use std::ffi::CString;
...
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::os::unix::io::FromRawFd;
 use std::path::Path;
 use util;
 
 #[cfg(all(lfs_support, target_os = "linux"))]
 use libc::{fstat64 as fstat, open64 as open, stat64 as stat_t};
 
 #[cfg(not(any(all(lfs_support, target_os = "linux"), target_os = "redox")))]
-use libc::{fstat, open, stat as stat_t};
```
Stebalien: I think this is broken for redox. I believe these two lines should be changed to

```rs
#[cfg(any(target_os = "linux", target_os = "emscripten", target_os ="l4re"))]
use libc::open64 as open;

#[cfg(not(any(target_os = "linux", target_os = "emscripten", target_os ="l4re")))]
use libc::open;
```

And then, I'd:

1. Move the `stat` definition for redox into the `same_dev_ino` definition (to keep everything together).
2. Import `libc::{fstat,stat as stat_t}` there.

---

This change also:

1. Removes the `lfs_support` attribute. That's an historical artifact that should be removed.
2. Adds LFS (large file support) on emscripten and l4re.

Stebalien:
I ended up making many of these changes in #82.

PR #64, opened 2018-07-17
Bump rand to 0.5

jonhoo:


Stebalien:
You'll need to bump the minimum rust version to 1.19 for the break with value.

P.S., hi.

jonhoo:
Ah, good catch! Fixed. `rand` should probably fix this to avoid forcing downstream crates to 1.19...

Also, hi! :D

jonhoo:
Turns out there are a bunch of other things that `rand` does that don't work in 1.18 either (like `impl Clone` for larger arrays), so I don't think fixing that is worthwhile. Bump to 1.19 it is!

jonhoo:
heh, yeah, those ones ^
I guess we have to bump to 1.22 judging by `rand`'s [`.travis.yml`](https://github.com/rust-lang-nursery/rand/blob/d625b180832910c4090506398689d4d60fb567f6/.travis.yml#L11)

Stebalien:
Meh. Rust is backwards compatible so I don't really mind relying on newer versions.

PR #67, opened 2018-09-22
Allow to disable hidden attribute on Windows

### LINKED ISSUE #66: Allow to create non hidden temporary files (on Windows)
https://github.com/Stebalien/tempfile/blob/f1273db441a5364185a52343eaba5b7127a6af36/src/file/imp/windows.rs#L27

I am using [winrt-rust](https://github.com/contextfree/winrt-rust) and I can't open temporary files with UWP APIs which have hidden bit set (E_ACCESSDENIED). It works if I stop on breakpoint and remove the attribute through explorer (I use NamedTempFile/TempPath).

nazar554:
Fixes #66

### DIFF
--- a/src/file/imp/windows.rs
+++ b/src/file/imp/windows.rs
@@ -24,8 +24,7 @@ const SHARE_MODE: DWORD = FILE_SHARE_DELETE
                         | FILE_SHARE_READ
                         | FILE_SHARE_WRITE;
 #[cfg_attr(irustfmt, rustfmt_skip)]
-const FLAGS: DWORD      = FILE_ATTRIBUTE_HIDDEN
-                        | FILE_ATTRIBUTE_TEMPORARY;
+const FLAGS: DWORD      = FILE_ATTRIBUTE_TEMPORARY;
 
 fn to_utf16(s: &Path) -> Vec<u16> {
     s.as_os_str()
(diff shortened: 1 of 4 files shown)


KodrAus:
Thanks @nazar554!

Before going ahead I'd like to explore whether there's a meaningful Unix implementation of `hidden`, and if not then the potential for platform-specific APIs for `tempfile::Builder` so we don't end up with a lot of features that are implicitly platform-specific.

nazar554:
Okay, the Unix build is failing anyway, I am going to fix it later.
In general on Unix (maybe also Redox) hidden files start with `.`, so maybe hidden can add it to `prefix`?
Also on FreeBSD (10+) and Mac OS X there is hidden flag set using `chflags +hidden`

KodrAus:
Hmm, having multiple builder methods able to mutate the same value can lead to some surprising behaviour.

The case I'm really trying to protect against here is picking arbitrary semantics on different platforms, and then having to support a suite of `hidden_*` methods over time with slightly different behaviour because those semantics don't make sense to everyone.

Just exploring some more design space, we could avoid adding a new method to the builder, and look at a way to modify the file flags using standard APIs _after_ the temporary file is created, using its `file_mut` or `into_file` methods to get access to it.

KodrAus:
It looks like we got a solution to this in #76.

Thanks again for the PR @nazar554!

PR #68, opened 2018-09-27
spooled temporary file

nlevitt:
For your consideration. Similar to python's `tempfile.SpooledTemporaryFile`: https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile


src/spooled.rs
```diff
@@ -0,0 +1,136 @@
...
+pub fn spooled_tempfile(max_size: usize) -> SpooledTempFile {
+    SpooledTempFile {
+        max_size: max_size,
+        cursor: Some(Cursor::new(Vec::new())),
+        file: None,
+    }
+}
+
+impl SpooledTempFile {
+    /// Returns true if the file has been rolled over to disk.
```
Stebalien: Can we have a `new(size)` constructor? I know it duplicates the `spooled_tempfile` function but that matches the rest of our code.

For context, we do this because `tempfile()` *doesn't* have a constructor because it doesn't have a type. Therefore, we create convenience functions for all tempfile types (for discoverability). However, it's (IMO) still idiomatic to have a constructor.

src/spooled.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// An object that behaves like a regular temporary file, but keeps data in
+/// memory until it reaches a configured size, at which point the data is
+/// written to a temporary file on disk, and further operations use the file
+/// on disk.
+#[derive(Debug)]
+pub struct SpooledTempFile {
+    max_size: usize,
+    cursor: Option<Cursor<Vec<u8>>>,
+    file: Option<File>,
```
Stebalien: Let's use an enum here. That is:

```rust
enum SpooledInner {
    InMemory(Cursor),
    File(File),
}
```

That way, we can't *have* any invalid cases. It should get rid of the panics.

src/spooled.rs
```diff
@@ -0,0 +1,136 @@
...
+impl SpooledTempFile {
+    /// Returns true if the file has been rolled over to disk.
+    pub fn rolled_over(&self) -> bool {
+        if let Some(ref _file) = self.file {
+            true
+        } else {
+            false
+        }
+    }
+}
```
Stebalien: We should probably provide an erroring `close()` method, for completeness (although I can't see why one would actually use it).

Stebalien: We should also provide a `set_len` (truncate) function. That one is significantly more useful.

Stebalien: A `rollover` method (like python's) would also be nice (although we can add that later).

nlevitt:
Thanks for the review! I implemented most of what you requested. I'm confused though about what `close()` is supposed to do. NamedTempfile has `close()`, which unlinks the file, but that doesn't apply here. `File` apparently has no `close()`, and `File` is sort of what we're emulating...

Stebalien:
> NamedTempfile has close(), which unlinks the file, but that doesn't apply here. File apparently has no close(), and File is sort of what we're emulating.

I've been writing too much go recently... I guess rust just uses `sync` and drop.

Stebalien:
Thanks!

PR #73, opened 2018-10-30
Attach a path to errors where possible

### LINKED ISSUE #71: Errors from creating files / directories should include the path that was attempted
I've been using `tempdir` in a project of mine but since that's deprecated in favor of this crate I figured I'd file this issue here.

While trying to track down [an issue in Firefox CI](https://bugzilla.mozilla.org/show_bug.cgi?id=1396845) I wound up patching `tempdir` to return the path it tried to create as part of the error message when it failed:
https://github.com/luser/tempdir/commit/bc045d5e96c85074da83efdf7ca512a2cfd5826b

Without that patch the error was just "The system cannot find the path specified. (os error 3)" which didn't give me much to work with. With that patch I instead got "Error creating temp directory "z:\\build\\build\\src\\c\\Users\\task_1537988825\\AppData\\Local\\Temp\\sccache.HQRR1j9faz6I": The system cannot find the path specified. (os error 3)" which was much more useful.

It would be nice if we could do this in the tempfile crate as well.

KodrAus:
Closes #71 

Adds an internal extension trait that lets us attach a `PathBuf` as context to an `io::Error` from temporary paths and directories when operations on them fail. It's not going to catch cases where we touch the underlying `File` itself, but should be a little more helpful for diagnostics.

cc @luser @Stebalien

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,40 @@
+use std::{error, io, fmt};
+use std::path::PathBuf;
+
+#[derive(Debug)]
+struct PathError {
+    path: PathBuf,
+    err: io::Error,
+}
+
+impl fmt::Display for PathError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{} at path {:?}", self.err, self.path)
+    }
+}
+
+impl error::Error for PathError {
+    fn description(&self) -> &str {
+        self.err.description()
+    }
+    
+    fn cause(&self) -> Option<&error::Error> {
+        self.err.cause()
+    }
+}
+
+pub(crate) trait IoErrorExt {
+    fn with_path<P>(self, path: P) -> Self where P: Into<PathBuf>;
+}
+
+impl IoErrorExt for io::Error {
+    fn with_path<P>(self, path: P) -> Self
+    where
+        P: Into<PathBuf>,
+    {
+        io::Error::new(self.kind(), PathError {
+            path: path.into(),
+            err: self,
+        })
+    }
+}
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -37,7 +39,9 @@ pub fn create_helper<F, R>(
 where
     F: Fn(PathBuf) -> io::Result<R>,
 {
-    for _ in 0..::NUM_RETRIES {
+    let num_retries = if random_len != 0 { ::NUM_RETRIES } else { 1 };
```
KodrAus: While testing this I was playing with attempting to create a file I could guarantee already existed by setting `random_len` to `0`. That's a bit of a footgun because you'll effectively try create the file 2147483648 times before giving up. Setting `random_len` to `0` is pretty silly, but we might want to just panic or use a much smaller number of retries if the available entropy is too small. I thought I'd leave this here so we could think about it more.

src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn cause(&self) -> Option<&error::Error> {
+        self.err.cause()
+    }
+}
+
+pub(crate) trait IoErrorExt {
+    fn with_path<P>(self, path: P) -> Self where P: Into<PathBuf>;
+}
+
+impl IoErrorExt for io::Error {
```
luser: If you wanted to make the code slightly less verbose you could impl this on `io::Result` instead, and just put the `map_err` in this function.

src/util.rs
Stebalien: One try sounds reasonable, nice catch! I can see users wanting to *explicitly* specify the path of a temporary file (if they just want a file that's automatically cleaned up).

Really, we should probably set NUM_RETRIES to something sane like 256 but we can do that later.

src/error.rs
KodrAus: Good idea :+1:

PR #75, opened 2018-11-26
Redox: Fix type definition for create_named

jackpot51:
Also, is it possible to bump the crates.io version after merging this?


jackpot51:
Thanks!

PR #78, opened 2019-02-07
chore: format

Stebalien:
(so my editor doesn't make random changes in every file I touch)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ rust:
   - nightly
   - beta
   - stable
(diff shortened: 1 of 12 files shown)


KodrAus:
From the build failure it looks like our minimum version might've been bumped under us?

Stebalien:
Bumped to 1.24... not sure which version we need.

Stebalien:
Yep: https://github.com/rust-lang/libc/issues/1232

Issue #40, opened 2018-02-07
Figure out how to deal with inaccessible parent directories


Stebalien:
So, unix really is lovely...

```sh
> mkdir -p /foo/bar
> cd /foo/bar
> chmod a-x /foo
> touch first
> touch /foo/bar/second
touch: cannot touch '/foo/bar/second': Permission denied
```

As written, *this* library has the same problem as the second case. We use the absolute path when creating temporary files instead of the relative path. I believe this can actually be an issue in practice with some jails.

Fixing this on *linux* is doable with `openat` and `unlinkat`. Fixing this on osx is possible using per-thread CWDs (at the cost of two extra syscalls to set/unset the thread-local CWD).

tbu-:
According to [What non-Linux Unixes support openat?](https://stackoverflow.com/a/32137401) from StackOverflow:
>[…], all major BSDs and Apple's OS X seem to support it as of today:
>* DragonFly since DragonFly 2.3.
>* FreeBSD since FreeBSD 8.0.
>* Linux since Linux 2.6.16 (for completeness).
>* NetBSD since NetBSD 7.0.
>* OpenBSD since OpenBSD 5.0.
>* OS X since OS X 10.10.
>* Solaris.

rbtcollins:
`fs_at` is a new crate I created as part of fixing the TOCTOU remove_dir_all inherited from the std::fs::remove_dir_all it was based on years ago; it uses NTCreateFile on windows and openat on unix - my intent for it is to be a medium-level crate: no sophisticated features or policy choices, just convenient access to 'at' style syscalls. If its not suitable for solving this bug, I'd like to know why so I can fix it.

Stebalien:
I'm tracking all 4.0 changes in https://github.com/Stebalien/tempfile/pull/327.

Stebalien:
Hm. Actually, I'm going to re-open this so we have a place to discuss it. To implement this:

1. `TempPath`, `NamedTempFile`, and `TempDir` would have to keep the parent directory open. However, even if we do that, I don't want to return relative paths when the user calls `.path()`. But if I return an absolute path, it'll probably break a good number of applications anyways.
2. `tempfile()` and `SpooledTempFile` would have to briefly open the parent directory when creating unnamed temporary files is impossible.

Maybe _just_ implement the latter?

nagisa:
Plenty of other ways in which a base directory name/path might start not working right:

1. Any component is a symbolic link that gets changed to something else;
[...]

PR #80, opened 2019-02-10
Use absolute paths

Stebalien:
When looking into #40, I realized we weren't actually _doing_ what I claimed. Except for never-named tempfiles on Linux and auto-removed temporary files on Windows, we need to use absolute path names to guard against changes in the current working directory.

Unfortunately, this is a **[breaking change]** on some unix-like OSs. It's possible to:

1. Chdir to a directory.
2. Chmod the parent directory to remove the search (`x`) permission.

This will break any uses of absolute path names. The solution is the one described in #40: open the current directory, use `openat` on the relative path, and then close the current directory. However:

1. This will only help with unnamed temporary files unless we're willing to _hold_ the file directory open (probably not the best idea).
2. This will require two filesystem extra syscalls (instead of one as with getcwd).

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -1,5 +1,6 @@
 #[cfg(not(target_os = "redox"))]
 use libc::{c_char, c_int, link, rename, unlink, O_CLOEXEC, O_CREAT, O_EXCL, O_RDWR};
+use std::env;
 use std::ffi::{CString, OsStr};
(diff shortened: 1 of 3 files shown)


Stebalien:
I snuck in two other small fixes while I was at it (second two commits).

Stebalien:
I snuck in two other small fixes while I was at it (second two commits).

PR #82, opened 2019-02-11
use modern stdlib features

Stebalien:
Based on #81.


KodrAus:
Looks like Windows isn't too happy though. I have a Windows box I can debug this on if you need?

Stebalien:
Turns out `create_new` conflicts with `access_mode` so I switched away from `access_mode`.

Stebalien:
(I've also squashed the commits as debugging-by-CI-and-rebasing was getting annoying)

>  I have a Windows box I can debug this on if you need?

Thanks! I _think_ I got it but, if this run doesn't work, I'm giving up...

PR #86, opened 2019-03-24
implement TempPath::keep

### LINKED ISSUE #84: NamedTempFile without delete
Hi, thanks for this library. It's very handy. I like that it's small and resposible for a single thing. But now I am about to ask a feature, which might not be intended to be implemented. :smile: 

I am about to implement a program which was written in Python before. Which was creating files in a given dir, with a given prefix and suffix. (But the file name contained random parts.) The files were temporary in a sense, that those were deleted by another process. The Python standard libray has the very same package (tempfile) which has the very similar type (NamedTemporaryFile). Which [allows](https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile) to disable the file removal when the handle is going out of scope.

To get this functionality, I need to reimplement most of what this crate does, but delete the `Drop` trait implementation. Or I can try to use the `persist` method, but I need to provide a random name somehow. (Not sure how to do that.)

What do you think: would that be okay to expose a `keep()` or `persist()` method on the `Builder` to ask for not to remove the file? Or how would you crate unique file names in a given directory (with given prefix and suffix)?

Stebalien:
fixes #84

### DIFF
--- a/src/file/imp/other.rs
+++ b/src/file/imp/other.rs
@@ -1,9 +1,12 @@
-use std::path::Path;
-use std::io;
 use std::fs::File;
+use std::io;
+use std::path::Path;
 
 fn not_supported<T>() -> io::Result<T> {
-    Err(io::Error::new(io::ErrorKind::Other, "operation not supported on this platform"))
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        "operation not supported on this platform",
+    ))
 }
 
 pub fn create_named(_path: &Path) -> io::Result<File> {
@@ -19,5 +22,9 @@ pub fn reopen(_file: &File, _path: &Path) -> io::Result<File> {
 }
 
 pub fn persist(_old_path: &Path, _new_path: &Path, _overwrite: bool) -> io::Result<()> {
+    not_supported()
+}
+
+pub fn keep(path: &Path) -> io::Result<()> {
     not_supported()
 }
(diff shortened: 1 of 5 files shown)


Stebalien:
cc @rizsotto

KodrAus:
Looks like the Windows build isn't happy:

```
error[E0599]: no method named `as_ptr` found for type `&std::path::Path` in the current scope
  --> src\file\imp\windows.rs:70:36
   |
70 |         if SetFileAttributesW(path.as_ptr(), FILE_ATTRIBUTE_NORMAL) == 0 {
   |                                    ^^^^^^
```

We might need to do something like:

```rust
let utf16_path = path
    .to_os_str()
    .encode_wide()
    .chain(Some(0))
    .collect::<Vec<u16>>();

if SetFileAttributesW(utf16_path.as_ptr(), ...
```

Which means allocating, but should be ok on this path?

rizsotto:
Thanks for doing this. My question is: would the `keep` method also available on the `tempfile::Builder` type too?

Stebalien:
@KodrAus debug by CI...

@rizsotto no. You'd have to call `keep` when you decide you want to keep the file. You could also just call `Builder::stuff().tempfile()?.keep()?`.

rizsotto:
Hey @Stebalien, thanks for your answer. I think the experience would be more convinient to have this method on the `Builder`. I also found hard to reason about why the method returns a `Result`.

Stebalien:
* It returns a `Result` because it may fail on windows (in windows, temporary files are marked temporary in the filesystem and have different persistence guarantees).
* This allows you to create the temporary file and then decide to keep it later. That is:

1. Create temporary file.
2. Open external process.
3. Do things...
4. Call `keep()`.

That way, if something goes wrong in 2/3, the temporary file sticks around.

---

However, we could (alternatively/also) add a `Builder::persistent_tempfile()` method (alternative to `Builder::tempfile()`) but that's strictly less useful.

PR #90, opened 2019-05-19
misc bug fixes

### LINKED ISSUE #25: PersistError is private

Stebalien:
* lint with clippy
* replace call to write with a call to write_all
* export PathPersistError (fixes #25)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -122,7 +122,7 @@ mod spooled;
 mod util;
 
 pub use dir::{tempdir, tempdir_in, TempDir};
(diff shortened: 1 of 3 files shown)


Stebalien:
(not going to bother with CR as these are all pretty simple fixes)

PR #92, opened 2019-06-03
docs: clarify the security concerns around named temporary files

Stebalien:


Stebalien:
@aztecrex does this help?

PR #94, opened 2019-06-30
Add an append flag to Builder

### LINKED ISSUE #93: Allow tempfile to be opened in append mode.
I've recently needed to be able to always write to the end of the file, and it looks like the crate doesn't provide a way to pass the O_APPEND flag to the underlying open call.

Overall, it almost looks like supporting std::fs::OpenOptions in Builder would help fix this, and maybe #30 in a generic fashion.

xavierd:
Fixes #93

### DIFF
--- a/src/file/imp/other.rs
+++ b/src/file/imp/other.rs
@@ -1,4 +1,4 @@
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io;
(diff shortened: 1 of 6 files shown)


Stebalien:
Thanks!

(test failures are deprecation notices/warnings, fixing in #95).

PR #98, opened 2019-06-30
bump rand

FauxFaux:
Tests are green for me on linux.

New `rand` has a significantly less disgusting dependency tree.


KodrAus:
Hi @FauxFaux :wave:

It looks like we need to bump our minimum supported version to update `rand`, which is done in #99 so we'll close this one in favour of #99.

Thanks for taking the time to work on this though!

PR #99, opened 2019-06-30
build: Bump rand to 0.7

lzutao:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tempfile"
-version = "3.0.9"
+version = "3.0.10"
```
lzutao: Or `3.1.0` ?

KodrAus: Since we're bumping our minimum supported version I think `3.1.0` would be preferable. What do you think @Stebalien?

Stebalien: Hm. Yeah, you're right. I already bumped the minimum in 3.0.9 but we should probably make minimum version bumps minor versions.

PR #104, opened 2019-07-18
build: Bump libc to 0.2

rizsotto:
Instead of specifying the `libc` crate dependency with PATCH version, I would suggest to use only the MINOR version.


Stebalien:
That patch version adds a feature tempfile depends on. Note: That version is short-hand for `^0.2.27` (any version in the range `[0.2.27, 0.3.0)`).

PR #105, opened 2019-07-18
build: Allow newer libc versions

rizsotto:
Follow up on #104 , using the caret form of the version definition.


Stebalien:
"0.2.27" is equivalent to "^0.2.27": https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

PR #106, opened 2019-08-04
Set MOVEFILE_WRITE_THROUGH flag on Windows

kornelski:
This is set by: https://github.com/untitaker/rust-atomicwrites/blob/02f7f492cc899d77a909cb83bff5ca698e4db20d/src/lib.rs#L207

MS documents it as:

> The function does not return until the file is actually moved on the disk.

When I replaced atomicwrites with tempfile in tantivy, the tests started failing, so I presume this flag is important.

### DIFF
--- a/src/file/imp/windows.rs
+++ b/src/file/imp/windows.rs
@@ -9,7 +9,7 @@ use std::{io, iter};
 use winapi::um::fileapi::SetFileAttributesW;
 use winapi::um::handleapi::INVALID_HANDLE_VALUE;
 use winapi::um::winbase::{MoveFileExW, ReOpenFile};
-use winapi::um::winbase::{FILE_FLAG_DELETE_ON_CLOSE, MOVEFILE_REPLACE_EXISTING};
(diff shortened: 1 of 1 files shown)


Stebalien:
Given that the standard library doesn't pass this flag in `rename`, I'm hesitant to do it here. Have you tried calling `sync_all()` on the file after moving it?

What are the symptoms you're seeing? This should only make a difference if the OS crashes before everything is written to disk. You may have a race condition elsewhere in your tests and this could be showing up due to changes in timing.

kornelski:
The test is here:

https://github.com/tantivy-search/tantivy/pull/612
https://ci.appveyor.com/project/fulmicoton/tantivy/builds/26452014

and the previous code was this:

https://github.com/untitaker/rust-atomicwrites/blob/a0db3c6ad80b834c50ee26fda51afd8c3d3f24d0/src/lib.rs#L204-L209

Stebalien:
This is either caused by a bug in Windows or a bug in your code. Unfortunately, I don't have time to dig into it to figure out what's going wrong.

The MOVEFILE_WRITE_THROUGH flag ensures that the move has been persisted _to disk_ before returning. Unless the OS crashes, it shouldn't be possible to observe the difference (unless you're somehow directly reading the contents of the disk).

PR #109, opened 2019-09-05
Fix minor documentation spelling issue

timvisee:


Stebalien:
Thanks but this is actually the correct spelling of "if and only if".

timvisee:
> Thanks but this is actually the correct spelling of "if and only if".

I didn't know. Thanks for the quick heads up!

Issue #110, opened 2019-09-06
clarification regarding the robustness of ``persist()``


phi-gamma:
Hi, I’m investigating what options there currently are regarding atomic
file writes. Assuming Linux, the common pattern requires that ``fsync(2)``
be called on the file fd before closing, then again on the fd of the containing
directory. AFAICS *tempfile* omits these fsync calls. Could you please
clarify whether this is by design? For potential users of the crate it would
be helpful to state the guarantees it provides in the readme.

Cf. this thread on the Rust forum: https://users.rust-lang.org/t/safe-way-to-write-file-to-disk/32099.

Stebalien:
This is by design.

In general, tempfile provides the same guarantees as the standard library and tries to provide the same granularity as the standard library. We _could_ sync files and directories on `persist()` but not everyone needs or wants that behavior. For example, a program might want to persist a bunch of files to a single directory, then sync each file, then sync the directory all at once.

But documenting the behavior can't hurt. Mind submitting a PR documenting this behavior on the `persist()` method?

phi-gamma:
Fair enough. Added a couple lines to the docs.

> For example, a program might want to persist a bunch of files to a single directory, then sync each file, then sync the directory all at once.

Currently the problem is syncing the directory – which is required by POSIX,
but not by some filesystems – due to the fact that *tempfile* operates on
paths, not file descriptors. Thus retroactively fsync()ing the directory is as
racy as the call to rename() is. To eliminate that race the API would need
extending to support passing the directory fd and then use renameat().

Stebalien:
Unfortunately, _that_ would require keeping the directory open, using an extra file descriptor. It also doesn't really help as I'm more concerned about the file being moved/deleted/replaced than the parent directory.

phi-gamma:
No problem, I’m not arguing. Just pointing out what currently prevents me
from using *tempfile*.

Stebalien:
Unless I'm missing something, it doesn't. You should be able to call persist, then sync, then sync the directory. Unless you're worried about someone coming by and renaming your _directory_ in which case you have other issues.

Stebalien:
> You can only sync a path which at that point may or may not refer
to the directory where the file has been moved. In most contexts I
agree that it is absolutely reasonable for an application to assume
that this is the case.

As I see it, there are two cases:

[...]

PR #111, opened 2019-09-09
Clarify the behavior of persist

phi-gamma:
Addresses issue #110

Add a note that persist() does not quite match the atomic rename
pattern that is customary on Linux.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -189,7 +189,10 @@ impl TempPath {
     /// If this method fails, it will return `self` in the resulting
     /// [`PathPersistError`].
     ///
-    /// Note: Temporary files cannot be persisted across filesystems.
+    /// Note: Temporary files cannot be persisted across filesystems. Also
+    /// neither the file contents nor the containing directory are
+    /// synchronized, so the update may not yet have reached the disk when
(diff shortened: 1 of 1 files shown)


src/file/mod.rs
```diff
@@ -189,7 +189,10 @@ impl TempPath {
     /// If this method fails, it will return `self` in the resulting
     /// [`PathPersistError`].
     ///
-    /// Note: Temporary files cannot be persisted across filesystems.
+    /// Note: Temporary files cannot be persisted across filesystems. Also
+    /// neither the file contents nor the containing directory are
+    /// synchronized, so the update may not yet have reached the disk when
+    /// `persist` returns.
```
Stebalien: Can you add a comment about how the user should resolve this (sync the file, then the parent directory).

phi-gamma: I refrained from doing so cause I’m not competent to give a portable answer.
On Linux, ``sync_all()`` on the file, then the directory is possible using just *std* (and
maybe *libc* for ``O_DIRECTORY``). Though how that maps to other OS which
tempfile supports, I have no idea.

Stebalien: SGTM.

PR #113, opened 2020-01-20
Change "iff" to "if" and remove apostrophe from "it's"

ConnorSkees:
Hello!

This is just a super small change to the docs.  
It looks like in 2 places "if" was typo'd as "iff."

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -243,7 +243,7 @@ impl TempPath {
         }
     }
 
-    /// Persist the temporary file at the target path iff no file exists there.
+    /// Persist the temporary file at the target path if and only if no file exists there.
     ///
     /// If a file exists at the target path, fail. If this method fails, it will
     /// return `self` in the resulting [`PathPersistError`].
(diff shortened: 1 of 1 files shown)


src/file/mod.rs
```diff
@@ -243,7 +243,7 @@ impl TempPath {
         }
     }
 
-    /// Persist the temporary file at the target path iff no file exists there.
+    /// Persist the temporary file at the target path if no file exists there.
```
Stebalien: This is actually intentional. Iff means "if and only if". However, you're not the first to point this out so it's probably better to just expand the phrase.

```suggestion
    /// Persist the temporary file at the target path if and only if no file exists there.
```

ConnorSkees: That makes sense! The rest of the documentation is really well written, so I thought a typo like that was really out of place. Thank you for the clarification :)

Stebalien: Well, the "it's" was definitely wrong and a bit embarrassing. Thanks for catching it!

(and iff needed to go anyways)

Issue #115, opened 2020-02-13
Got bitten by Drop impl running to early when passing TempDir to fn(p: impl AsRef<Path>)


robinhundt:
Yesterday i got badly bitten by using `tempdir()` in conjunction with `std::process::Command::current_dir(...)` . My initial code looked like this:
```rust
let tmp_dir = tempdir();
let output = Command::new(bin)
    .current_dir(tmp_dir)
    .output()?;  
```
with `bin` doing file operations inside `tmp_dir`. Unfortunately this did not work, and I received an error along the lines of `Os Error 2: No such file or directory`.  
After beating my head against the screen for some time i thought "maybe the tmp_dir is dropped to early and the directory deleted before `bin` is run?" so i changed
```rust
    .current_dir(tmp_dir)
```
to 
```rust
    .current_dir(&tmp_dir)
```

and my program worked just as intended.  
Like the issue title says, this was due to the signature of 
```rust
pub fn current_dir<P: AsRef<Path>>(&mut self, dir: P) -> &mut Command
```
Since `TempDir` implements `AsRef<Path>` i thought that moving it in would be the right thing to do (i didn't need to access the tmp_dir later in the method after constructing the command). But this lead to `Drop` impl of `TempDir` being run after `current_dir` finished but before the execution of `bin`.

This kind of mistake can quickly happen, especially since many APIs dealing with Paths are generic over `AsRef<Path>`.  
**My question is** if this pitfall should be pointed out in the documentation and if so, where the best place would be to do this? I'd like to start contributing to some rust crates and this seems like a really easy thing to do, but which could have saved me some time and nerves :D

I was also thinking if it's possible to prevent this mistake, maybe by changing the `AsRef<Path>` impl to:
```rust
impl AsRef<Path> for &TempDir
```
but this  would be a breaking change.

Edit: I assume all of this applies to `TempFile` as well.

Stebalien:
This is an interesting issue. You're right that changing the impl to `&TempDir` (and `&mut TempDir`) would fix this, the real question is, as you say, how much would it break. Implementing `AsRef<Path>` on `TempDir` by value is, in theory at least, a useful feature: structs can _store_ the reference to the temporary directory without caring about what it actually is, and have it cleaned up when they're done with it.

On the other hand, nearly all AsRef users immediately call `as_ref` and convert the path into an owned path if they actually need to store it. I've never actually seen a struct that's generic over something implementing `AsRef<Path>`.

[...]

PR #116, opened 2020-02-15
Add docs pointing out early dop pitfalls

robinhundt:
This commit adds a section to the `src/libs.rs` docs
pointing out an easily encounterable error when
ussing `tempdir()` with apis that are generic over
`AsRef<Path>`, like `Command::current_dir`, resulting
in the directory being dropped to early (As pointed out in #115).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,39 @@
 //! rely on file paths for _some_ operations. See the security documentation on
 //! the `NamedTempFile` type for more information.
 //!
+//! ## Early drop pitfall
+//! Because `TempDir` and `NamedTempFile` rely on their destructors for cleanup, this can lead
+//! to an unexpected early removal of the directory/file, usually when working with APIs which are
+//! generic over `AsRef<Path>`. Consider the following example:
+//!
+//! ```
+//! # use tempfile::tempdir;
+//! # use std::io;
+//! # use std::process::Command;
+//! # fn main() {
+//! #     if let Err(_) = run() {
+//! #         ::std::process::exit(1);
+//! #     }
+//! # }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -27,6 +27,39 @@
 //! rely on file paths for _some_ operations. See the security documentation on
 //! the `NamedTempFile` type for more information.
 //!
+//! ## Early drop pitfall
+//! Because `TempDir` and `NamedTempFile` rely on their destructors for cleanup, this can lead
+//! to an unexpected early removal of the directory/file, usually when working with APIs which are
+//! generic over `AsRef<Path>`. Consider the following example:
+//!
+//! ```
```
robinhundt: I'm not sure if this example should be kept as a doc test or maybe markes as `no_run` since it would fail on platforms where touch is not available.

Stebalien: Let's mark it `no_run` for now.

src/lib.rs
```diff
@@ -27,6 +27,39 @@
...
+//!
+//! # Ok(())
+//! # }
+//! ```
+//! This works because a reference to `temp_dir` is passed to `current_dir`, resulting in the
+//! destructor of `temp_dir` being run after the `Command` has finished execution. Moving the
+//! `TempDir` into the `current_dir` call would result in the `TempDir` being converted into
+//! an internal representation, with the original value being dropped and the directory thus
+//! being deleted, before the command can be executed.
+//! The `touch` command would fail with an `No such file or directory` error.
```
Stebalien: Markdown needs blank lines between paragraphs. Really, we should have blank lines between everything (headings, code blocks, etc.).

Stebalien:
Continued in #165.

PR #117, opened 2020-04-18
Fix two typos

bowlofeggs:
desctructor → destructor

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


Stebalien:
Thanks!

PR #124, opened 2020-08-25
Use more direct code than mem::replace

rbtcollins:
The mem::replace idiom was unneeded, so replace it with more direct
code.

This fixes the lint issues breaking the build on master.


Stebalien:
Thanks!

Issue #120, opened 2020-06-25
remove_dir_all caret dependency is incompatible with tempfile MSRV


str4d:
The recent 0.5.3 release of the `remove_dir_all` crate bumped its effective MSRV to 1.40.0, and as now documented in https://github.com/XAMPPRocky/remove_dir_all/issues/21#issuecomment-649295073, the crate only supports the latest Rust version. However, the `remove_dir_all` crate is specified as a (default) caret dependency of `tempfile`:

https://github.com/Stebalien/tempfile/blob/6fe7f2c4b270c9766580010ba842e59bda9a1cf3/Cargo.toml#L22

This automatically selects 0.5.3, breaking `tempfile`'s MSRV of 1.32.0.

`tempfile` should either pin with `=0.5.2` until its MSRV reaches 1.40.0 (and always pin specific versions going forward, as caret dependencies may similarly break in future), or remove the `remove_dir_all` dependency.

Stebalien:
It's about time to bump the MSRV anyways. 1.40 is 6 months old at this point.

PR #125, opened 2020-09-26
fix build error from remove_dir_all version

sbosnick:
The release of remove_dir_all version 0.5.3 increased the implicit MSRV of that crate to 1.40.0. The policy of the author of that crate is to support only the current stable version of Rust and to not treat a change in the implicit MSRV as a breaking change (see #XAMPPRocky/remove_dir_all#21). The author of that crate recommends pinning to an earlier version to continue to support an earlier MSRV.

Increase the MSRV in this crate to 1.40.0 to allow it to compile with remove_dir_all version 0.5.3. Also pin remove_dir_all to version 0.5.3 to avoid a caret dependency that could break this crate's MSRV unexpectedly based on future changes to remove_dir_all. 

Closes: #120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rust:
   - nightly
   - beta
   - stable
-  - 1.32.0
+  - 1.40.0
 os:
   - linux
   - osx
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ description = "A library for managing temporary files and directories."
 [dependencies]
 cfg-if = "0.1"
 rand = "0.7"
-remove_dir_all = "0.5"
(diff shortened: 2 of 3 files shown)


Stebalien:
While I try to support older rust versions, within reason, I also don't consider version updates to be breaking changes.

If it becomes an _actual_ problem (e.g., MSRV bumps to a release less than 6 months old), I'm happy to introduce a maximum version to fix any fallout. But without that, I'd rather not pin a dependency (I'm not going to monitor it for releases and update the pinned version).

sbosnick:
I see that you beat me to this change with #126. I am closing this pull request as it is now just a duplicate of that one.

PR #126, opened 2020-09-26
bump msrv to 1.40.0

### LINKED ISSUE #120: remove_dir_all caret dependency is incompatible with tempfile MSRV

Stebalien:
Required by remove-dir-all.

fixes #120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rust:
   - nightly
   - beta
   - stable
-  - 1.32.0
+  - 1.40.0
 os:
   - linux
   - osx
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ patterns and surprisingly difficult to implement securely).
 Usage
 -----
 
-Minimum required Rust version: 1.32.0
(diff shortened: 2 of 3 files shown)


dekellum:
So next release of tempfile will be 3.2.0, given this change?   (Recent MSRV bumps have been done with a MINOR release.)

PR #128, opened 2020-09-28
Add builder() shortcut method to TempDir / TempFile

dbrgn:
In the source code, using `TempDir::builder()` is more expressive than `Builder::new()`.

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -263,6 +263,13 @@ impl TempDir {
         Builder::new().tempdir_in(dir)
     }
 
+    /// Shortcut for [`Builder::new()`].
+    ///
+    /// [`Builder::new()`]: struct.Builder.html#method.new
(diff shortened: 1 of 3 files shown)


Stebalien:
This will lead to a pretty confusing API: `TempDir::builder().tempfile()`.

dbrgn:
Yes, that is the tradeoff. It's up to you if you want that or not. (However, I'd like to note that we're already doing `tempfile::Builder::new().tempdir()` right now, which might look equally confusing :stuck_out_tongue:)

Stebalien:
Well, the crate name is tempfile. I guess `temp` would be a better crate name, but tempdir support was merged after the fact...

Choosing between a new potential source of confusion and an existing source of potential confusion, I'd rather keep the existing one. You're right that both are a bit confusing, but the current state is at least logically consistent.

dbrgn:
No problem :slightly_smiling_face:

PR #130, opened 2020-10-12
Update to cfg-if 1.0

mbrubeck:
The version was bumped for alexcrichton/cfg-if#37, which removed some inconsistent syntax which has no effect on this crate.  There are no other changes compared to the previous version.


Stebalien (approved):

Stebalien:
Thanks!

Issue #132, opened 2020-10-22
Cannot persist file to separate disk drive


qwerty01:
On Windows, when persisting a temporary file to a drive other than the location of the temp folder, the following error is returned:
```
Os { code: 17, kind: Other, message: "The system cannot move the file to a different disk drive." }
```
This seems to occur because `MoveFileEx` requires the `MOVEFILE_COPY_ALLOWED` for any moves that require copying file data.

qwerty01:
As mentioned in #133, crossing drive boundaries breaks the condition that persisting is an atomic operation.  In order to avoid drive boundaries, the temporary file must be made on the same drive, which can be accomplished with `NamedTempFile::new_in`, allowing you to specify a directory to create the temporary file in.

PR #133, opened 2020-10-22
Allows copying file data when persisting

qwerty01:
Allows persisting temp files to copy on move, if required. Solves issue #132


Stebalien:
Unfortunately, that's not atomic which is why persisting across filesystems is not supported (https://docs.rs/tempfile/3.1.0/tempfile/struct.TempPath.html#method.persist). This won't work on any platform.

If you plan on using persist, you're going to need to make your temporary files in the target directory (where you want to persist them).

qwerty01:
Ah that's true, I don't think there's any way around that, so I'll just close this request.

Issue #112, opened 2019-11-09
Choose a writable temp dir please


gilescope:
Reading the docs for the windows temp dir selection, it goes through several fallbacks before it selects a dir:

https://docs.microsoft.com/en-gb/windows/win32/api/fileapi/nf-fileapi-gettemppatha

Would it be possible for this crate to check that the dir selected is writable by the current user before selecting it? ( See https://github.com/rust-lang/cargo/issues/4350 )

KodrAus:
Ah that's a bit unfortunate! We could write our own alternative to `env::temp_dir`, like `writable_temp_dir`, that on Windows follows through the same environment variables, and on Linux just forwards to `env::temp_dir`.

Would you like to submit a PR that does something like this?

Stebalien:
I'm not sure this is the right behavior. The docs imply that one should:

1. Call `GetTempPath` to get the location where temporary files _should_ be written.
2. Check to make sure that this directory is accessible.

That's effectively what we're doing by trying to actually create a file/directory.

The docs _don't_ say that one should check if the paths specified by `TMP`, `TEMP`, and `USERPROFILE` are writable in-turn.

From what I can tell, specifying a `TMP` that isn't writable is a misconfiguration. I'd rather fail early than work around it.

hunger:
I would appreciate having a way to override the temporary directory that cargo uses somewhere in the configuration! I am on Linux, and /tmp is writeable. The build starts, but it runs out of memory whenever I a bigger crate is built!

/tmp is a tmpfs in the distribution I use, and limited to 8G!

IIRC the recommended place to use for bigger temporary data is /var/tmp on Linux, that would definitely held for my case.

I would love to be able to just set a tmpdir in the cargo configuration file! I do not want to globally set TMPDIR and I keep forgetting to do so whenever I run cargo.

Stebalien:
@hunger please ask the cargo team for an override. That's not something this library can help with (and not related to this issue).

hunger:
@Stebalien : Oh, sorry, I got here by following a link from the cargo bug tracker and had not noticed that I was in a different project now!

spoelstraethan:
Writable isn't the only issue, if you can write files there but can't execute them, you are still going to get build failures. Not sure if that is a separate issue or just another unit/integration test to use against the code changes.

PR #138, opened 2021-02-01
Add [limited] support for WASI

RReverser:
WASI doesn't have a concept of system-wide temporary directory, but at least there is no reason why `new_in` methods with explicit paths should fail.

This PR implements support for such methods.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ cfg-if = "1"
 rand = "0.8"
 remove_dir_all = "0.5"
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 libc = "0.2.27"
 
 [target.'cfg(windows)'.dependencies.winapi]
@@ -34,3 +34,6 @@ features = [
 
 [target.'cfg(target_os = "redox")'.dependencies]
 redox_syscall = "0.2"
+
+[features]
+nightly = []
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -2,10 +2,18 @@ use std::env;
 use std::ffi::{CString, OsStr};
 use std::fs::{self, File, OpenOptions};
 use std::io;
-use std::os::unix::ffi::OsStrExt;
-use std::os::unix::fs::{MetadataExt, OpenOptionsExt};
-use std::path::Path;
+cfg_if! {
+    if #[cfg(not(target_os = "wasi"))] {
+        use std::os::unix::ffi::OsStrExt;
+        use std::os::unix::fs::{MetadataExt, OpenOptionsExt};
+    } else {
+        use std::os::wasi::ffi::OsStrExt;
+        #[cfg(feature = "nightly")]
+        use std::os::wasi::fs::MetadataExt;
+    }
+}
 use crate::util;
+use std::path::Path;
 
 #[cfg(not(target_os = "redox"))]
 use libc::{c_char, c_int, link, rename, unlink};
@@ -33,12 +41,14 @@ pub fn cstr(path: &Path) -> io::Result<CString> {
 }
 
 pub fn create_named(path: &Path, open_options: &mut OpenOptions) -> io::Result<File> {
-    open_options
-        .read(true)
-        .write(true)
-        .create_new(true)
-        .mode(0o600)
-        .open(path)
+    open_options.read(true).write(true).create_new(true);
+
+    #[cfg(not(target_os = "wasi"))]
+    {
+        open_options.mode(0o600);
+    }
+
+    open_options.open(path)
 }
 
 fn create_unlinked(path: &Path) -> io::Result<File> {
@@ -94,11 +104,20 @@ pub fn reopen(file: &File, path: &Path) -> io::Result<File> {
     let new_file = OpenOptions::new().read(true).write(true).open(path)?;
     let old_meta = file.metadata()?;
     let new_meta = new_file.metadata()?;
-    if old_meta.dev() != new_meta.dev() || old_meta.ino() != new_meta.ino() {
-        return Err(io::Error::new(
-            io::ErrorKind::NotFound,
-            "original tempfile has been replaced",
-        ));
+    cfg_if! {
+        if #[cfg(any(not(target_os = "wasi"), feature = "nightly"))] {
+            if old_meta.dev() != new_meta.dev() || old_meta.ino() != new_meta.ino() {
+                return Err(io::Error::new(
+                    io::ErrorKind::NotFound,
+                    "original tempfile has been replaced",
+                ));
+            }
+        } else {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "this operation is supported on WASI only on nightly Rust (with `nightly` feature enabled)"
+            ));
+        }
     }
     Ok(new_file)
 }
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -34,3 +34,6 @@ features = [
 
 [target.'cfg(target_os = "redox")'.dependencies]
 redox_syscall = "0.2"
+
+[features]
+nightly = []
```
Stebalien, line 39: How about using https://github.com/dtolnay/rustversion?

src/file/imp/unix.rs
```diff
@@ -94,11 +104,20 @@ pub fn reopen(file: &File, path: &Path) -> io::Result<File> {
     let new_file = OpenOptions::new().read(true).write(true).open(path)?;
     let old_meta = file.metadata()?;
     let new_meta = new_file.metadata()?;
-    if old_meta.dev() != new_meta.dev() || old_meta.ino() != new_meta.ino() {
-        return Err(io::Error::new(
-            io::ErrorKind::NotFound,
-            "original tempfile has been replaced",
-        ));
+    cfg_if! {
+        if #[cfg(any(not(target_os = "wasi"), feature = "nightly"))] {
```
Stebalien, line 108: Let's do this at the function level. That is, define one function that always returns an error, and one function that actually works.

Stebalien: Also, does this actually _work_ on wasi? I'd expect dev/ino to return garbage.

Stebalien (changes requested):
Thanks!

Stebalien:
Also, if there _is_ a good way to test this that would be nice. I've run introduced an unfortunate number of bugs in platforms not tested in CI.

src/file/imp/unix.rs
RReverser: Yeah it does, dev/ino are parts of WASI API and are propagated from WASI runtime. https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-filestat-struct

RReverser:
> Also, if there is a good way to test this that would be nice. I've run introduced an unfortunate number of bugs in platforms not tested in CI.

So that's kinda the problem - testing itself is easy, but I was surprised to find that all the currently written tests use methods based on `std::env::temp_dir()`.

I can write a test or two specifically for `new_in` + WASI, but then I'd still need to somehow disable all the other existing tests, as they definitely fail on WASI, and things get ugly plus coverage is pretty poor...

Cargo.toml
RReverser: Yeah fine by me if that's your preference.

Stebalien:
> So that's kinda the problem - testing itself is easy, but I was surprised to find that all the currently written tests use methods based on std::env::temp_dir().

Take a look in the `tests` directory.

> I can write a test or two specifically for new_in + WASI, but then I'd still need to somehow disable all the other existing tests, as they definitely fail on WASI, and things get ugly plus coverage stays pretty poor anyway, so I'm not sure if it's worth it...

You should be able to disable the tempdir tests using a `#![cfg(...)]` at the top of the test file, IIRC (I haven't done this kind of thing in a while).

Honestly, coverage isn't really something I'm too concerned about here. I just want to make sure it compiles and basically works. Compiles is probably a high enough bar if you can get that tested in CI.

Cargo.toml
Stebalien: Actually..., never mind. Thinking through this, I'd rather someone _explicitly_ opt-in to this behavior if they need it rather than be surprised that something works on nightly but not elsewhere.

RReverser:
> Take a look in the `tests` directory.

I did, but I guess I missed `test_tempdir`: https://github.com/Stebalien/tempfile/blob/57072e43a24c5d683e7e203da8e94bc91fb92830/tests/tempdir.rs#L44-L52

Aside from that, all the other tests seem to rely on `std::env::temp_dir()` either directly or indirectly.

> Compiles is probably a high enough bar if you can get that tested in CI.

Ah, that should be trivial. I'll look into that tomorrow.

Stebalien:
> Aside from that, all the other tests seem to rely on std::env::temp_dir() either directly or indirectly.

Oh, sorry, I misread. I thought you were saying that only temporary directories were tested.

Yeah, I see the issue. Eventually it would probably be a good idea to just pick a sane default for `std::env::temp_dir()`, but just compiling in CI is likely enough for now.

RReverser:
Ok I've added two jobs - one for WASI on Rust 1.40.0 on Linux, and one for WASI on nightly Rust with nightly feature on Linux.

I believe those should be sufficient, as they cover MSRV as well as nightly build, and we don't want to test all possible combinations as that quickly explodes number of jobs and significantly increases CI time for little benefit.

One more thing I noticed is that AppVeyor shows up on Github PR checks and commits, but Travis doesn't - perhaps integration is not correctly enabled in repo settings?

RReverser:
> Eventually it would probably be a good idea to just pick a sane default for `std::env::temp_dir()`, but just compiling in CI is likely enough for now.

As for this - I was thinking about potential solutions, and I wonder if a better option would be to expose API to customize root temp dir instead? It would not only avoid hardcoding a default that might not match WASI user's expectations, but also potentially help users on other systems too in case they want to choose a custom root temporary dir (e.g. this could be used as a workaround in https://github.com/Stebalien/tempfile/issues/112).

I'm not ready to take on this in a follow-up PR, just sharing an idea :)

Stebalien:
> As for this - I was thinking about potential solutions, and I wonder if a better option would be to expose API to customize root temp dir instead? It would not only avoid hardcoding a default that might not match WASI user's expectations, but also potentially help users on other systems too in case they want to choose a custom root temporary dir (e.g. this could be used as a workaround in #112).

Ideally, they'd set the default through the system itself (e.g., TMPDIR). I prefer to avoid providing global "setters" as they tend to be abused.

Stebalien (approved):
Thanks!

RReverser:
> Ideally, they'd set the default through the system itself (e.g., TMPDIR). 

Yeah that's an option too, but still a little concerning because WASI intentionally _doesn't_ try to emulate POSIX, and intentionally tries not to rely on any environment variables.

It's still possible for separate crates to choose to rely on those, of course, but a bit worried that it might create inconsistency between parts of the ecosystem...

Maybe a crate-specific env var like `TEMPFILE_RS_DIR` would be fine though.

RReverser:
Although... maybe it wouldn't be too bad. I guess `TMPDIR` is a fairly explicit signal, especially if documented in the crate.

Stebalien:
At the moment, I think it's best to wait. I assume the WASI runtime environment will eventually pick _something_.

RReverser:
> One more thing I noticed is that AppVeyor shows up on Github PR checks and commits, but Travis doesn't - perhaps integration is not correctly enabled in repo settings?

Looks like this made it easy to miss one compilation issue with WASI in latest commit... I'll make a follow-up fix, but I think integration to see checks in Github could be helpful anyway :) https://travis-ci.org/github/Stebalien/tempfile/builds/757241191

Stebalien:
Unfortunately, travis is switching from travis-ci.org to travis-ci.com. Along with that, they're demanding access to _all_ repos, public and private.

In other news, I'm going to be looking for a new CI service.

RReverser:
FWIW I've been starting new projects as well as migrating some old ones to Github Actions, and it's been a breeze so far. A lot faster than Travis, too.

RReverser:
Oh, and they have Windows support, definitely faster than AppVeyor as well :)

Stebalien:
Thanks!

PR #141, opened 2021-03-24
use rand's SmallRng for random number generation

mcginty:
#119 seems to have not been updated, so here's another take at it, trying to keep the diff as small as possible.

This approach stores a `SmallRng` as a thread-local variable, basically [directly copying the way `rand` does its own `thread_rng()`](https://github.com/rust-random/rand/blob/master/src/rngs/thread.rs) so as to avoid any issues or incompatibilities.

I opted not to further change the way that the `OsString` is built since that feels like it may be better in another PR.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,12 +1,22 @@
-use rand::distributions::Alphanumeric;
-use rand::{self, Rng};
+use rand::{self, Rng, SeedableRng};
+use rand::{distributions::Alphanumeric, rngs::SmallRng};
 use std::ffi::{OsStr, OsString};
-use std::path::{Path, PathBuf};
+use std::thread_local;
+use std::{
+    cell::UnsafeCell,
+    path::{Path, PathBuf},
+    rc::Rc,
+};
 use std::{io, str};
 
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -1,20 +1,30 @@
...
+    cell::UnsafeCell,
+    path::{Path, PathBuf},
+    rc::Rc,
+};
 use std::{io, str};
 
 use crate::error::IoResultExt;
 
+thread_local! {
+    static THREAD_RNG_KEY: Rc<UnsafeCell<SmallRng>> = Rc::new(UnsafeCell::new(SmallRng::from_entropy()));
```
Stebalien, line 15: We can drop the Rc, right? Just use `THREAD_RNG_KEY.with(|rng| use rng...)`.

Stebalien (changes requested):
One small nit, but otherwise LGTM.

Stebalien (approved):

mcginty:
@Stebalien sorry I totally spaced on getting those changes in, thanks so much for merging!

PR #144, opened 2021-04-08
Ensure that README code example is up to date

GuillaumeGomez:
This provide a simple way to check that your README example is up-to-date.


Stebalien (approved):
Thanks!

Issue #137, opened 2021-01-07
Make permissions and ownership? configurable while persisting


xaverdh:
Related issues: #30, #114

It might not be possible to support on all platforms, but it would still be nice to have it at least on unix. This issue is specifically about permissions / ownership of the persisted file rather than the temporary file, but I would also be quite happy with an option to configure permissions on the temporary, if they get carried over to the persisted file.

My use case is to atomically write files with given permissions / ownership by first writing to a temporary file and then persisting it (relying on rename likely being atomic), so just changing these after the fact will slightly defeat the purpose..

xaverdh:
Implementation wise, I think the most realistic approach is to make permissions on the temporary file configurable and then make persist keep those. Ownership could be configured through the `Builder`, but would only come into effect when persisting.

On Linux this would correspond to passing `O_TMPFILE`, but without `O_EXCL` and creating it with given mode. Then just before persisting it with `linkat` you would call `fchmod` to adjust ownership.
I don't know how well this fits with other platforms though..

Stebalien:
As far as I know, persist _should_ keep the permissions. So you should be able to:

1. Create a named temporary file.
2. Chmod the temporary file.
3. Call persist.

> On Linux this would correspond to passing O_TMPFILE, but without O_EXCL and creating it with given mode. Then just before persisting it with linkat you would call fchmod to adjust ownership.

This library does not support persisting temporary files created with `O_TMPFILE`, unfortunately. I'd like to implement this, but I have a policy of only providing features that are truly cross-platform (linux, windows, macos, freebsd), including sharing equivalent security guarantees.

D1plo1d:
> This library does not support persisting temporary files created with `O_TMPFILE`, unfortunately. I'd like to implement this, but I have a policy of only providing features that are truly cross-platform (linux, windows, macos, freebsd), including sharing equivalent security guarantees.

Would it be possible to remove the `O_EXCL` flag entirely? Is there a security difference from eg. windows or some other OS if it is removed? It appears that according to the [gnu libc docs](https://www.gnu.org/software/libc/manual/html_node/Open_002dtime-Flags.html) `O_EXCL | O_TMPFILE` only serve to prevent persisting tempfiles and it's removal would allow developers to persist tempfiles externally to `tempfile.rs` using `nix`.

For example without `O_EXCL` the following works:

```rust
tmpfile::tmpfile();
let tmp_file_path = format!("/proc/self/fd/{}", tmp_file.as_raw_fd());

[...]

PR #148, opened 2021-05-15
Remove `O_EXCL` to allow for tmpfile persistence on Linux

D1plo1d:
Per #137 this PR remove the `O_EXCL` flag on Linux to enable Linux users to persist temp files without copying their contents.

### Example Usage

```rust
// Note: Persisting tmp files will not work if the tmp directory and
// permanent directory are not on the same file system
let file = tmpfile::tmpfile_in(some_directory_on_the_same_fs);

// TODO: Write to the tmp file like normal here
// ...

// Use /proc/self/fd/ to get a file path for the unnamed temp file
let tmp_file_path = format!("/proc/self/fd/{}", file.as_raw_fd());

// Persist the temp file to a new named path
nix::unistd::linkat(
    None,
    &tmp_file_path[..],
    None,
    &file_path[..],
    nix::unistd::LinkatFlags::SymlinkFollow,
)?;
```

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -70,11 +70,11 @@ fn create_unlinked(path: &Path) -> io::Result<File> {
 
 #[cfg(target_os = "linux")]
 pub fn create(dir: &Path) -> io::Result<File> {
-    use libc::{EISDIR, ENOENT, EOPNOTSUPP, O_EXCL, O_TMPFILE};
+    use libc::{EISDIR, ENOENT, EOPNOTSUPP, O_TMPFILE};
     OpenOptions::new()
         .read(true)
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

Stebalien:
Thanks!

D1plo1d:
Np! 😄

dekellum:
Doesn't removing O_EXCL also have some security implications?  AFAICT, the flag is used to error on an existing file with same name and avoids following symlinks.

Stebalien:
Yes. However, that's not an issue here as we're creating an unnamed file which, by definition, can't conflict with existing files (doesnt' have a name).

Issue #129, opened 2020-10-06
Depend on remove_dir_all 0.6


dvtomas:
I'm removing duplicate dependencies from my project, and remove_dir_all 0.5 (tempfile) vs 0.6 (others) is one of them

Stebalien:
Note: This is currently blocked because `remove_all_dirs` 0.6+ adds a dependency on rayon, and I'm strongly against having a temporary file library launch new threads.

messense:
Is `remove_dir_all` still needed? Cargo recently removed it and `std::fs::remove_dir_all` shows better performance: https://github.com/rust-lang/cargo/pull/11333

Xaeroxe:
Resolved in #207. Also, `remove_dir_all` no longer requires `rayon`.

stanislav-tkach:
By the way, there is a race condition in old (<0.8.0) versions of the `remove_dir_all` library, so it would be nice to either update or remove this crate.

Xaeroxe:
Changed my mind, switched to removing it in #208

adumbidiot:
Was https://github.com/rust-lang/rust/issues/29497 worked around?

PR #149, opened 2021-05-18
update remove_all_dirs to 0.7

jbtrystram:
Bump remove all dirs to a more recent version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ description = "A library for managing temporary files and directories."
 [dependencies]
 cfg-if = "1"
 rand = { version = "0.8", features = ["small_rng", "getrandom"], default_features = false }
-remove_dir_all = "0.5"
+remove_dir_all = "0.7"
 
 [target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 libc = "0.2.27"


Stebalien:
See https://github.com/Stebalien/tempfile/issues/129 (sorry, should have left a comment here earlier). I don't like depending on older versions, but I _definitely_ don't want a temporary file library spinning off threads, depending on rayon, etc.

jbtrystram:
Thanks for the explanation

apiraino:
hi @Stebalien I'd love to learn more on this topic.

I notice the `remove_all_dirs` crate maintainers declare that their crate behaves so that "on non-Windows this is a re-export of `std::fs::remove_dir_all`" and only on Windows target the `rayon` dependency adds concurrent files deletion. I think this is [their line of thought](https://github.com/XAMPPRocky/remove_dir_all/blob/c5c5d11d1634e3612b1841496b4b298f4230bea9/src/fs.rs#L26-L58) on why are using `rayon`.  Towards the end they mention a possible race condition on concurrent files deletion.  

I'd be happy if you could share your thoughts on why using a concurrent approach is not great. I'm not questioning your choice, just interested in more technical details.

If `remove_dir_all=0.5` is feature-complete in the context of the `tempfile` crate and there are no other impellent reasons to push for an upgrade, as a user I'm fine with this crate holding the dependency upgrade, otherwise did you evaluate if this warrants dropping `remove_all_dirs` for something else?

Would it be possible to use two versions of the `remove_all_dirs` dependency? Let's say 0.5 on Windows and the latest on all the other platforms, example:
```
[target.'cfg(any(unix))'.dependencies]
remove_dir_all = "0.7"
[target.'cfg(any(windows))'.dependencies]
remove_dir_all = "0.5"
```
Would that be a nightmare from the maintenaince point of view?

Final question (perhaps related): is this the reason why `tempfile` stays synchronous and does not leverage the [Tokio](https://docs.rs/tokio/1.6.1/tokio/fs/struct.File.html)/[Async-std](https://docs.rs/async-std/1.9.0/async_std/fs/struct.File.html) runtime for async operations?

thank you so much :-)

Stebalien:
> I'd be happy if you could share your thoughts on why using a concurrent approach is not great. I'm not questioning your choice, just interested in more technical details.

1. It's a large, complex dependency tree. Ideally adding `tempfile` to your project shouldn't add too much weight.
2. The principle of least surprise. I'd be very surprised if a temporary file library decided to spin up a thread pool on drop. I'd likely make safety assumptions assuming this can't happen.

> I'd be happy if you could share your thoughts on why using a concurrent approach is not great. I'm not questioning your choice, just interested in more technical details.

Unfortunately, I'm not aware of any reasonable alternatives at the moment. I'd love to upgrade, but I'd need two things:

1. A feature to disable pulling in the rayon dependency (issue 1).
2. A variant of `remove_dir_all` that doesn't spin up a thread pool.

With that, I'd be happy to add a `parallel_delete` option to the tempdir builder to actually make use of the parallel deletion feature. I just want it to be opt-in.

> Would it be possible to use two versions of the remove_dir_all dependency?

Does that really buy us anything? That seems like it's more likely to cause confusion than actually solve anything.

> Final question (perhaps related): is this the reason why tempfile stays synchronous and does not leverage the Tokio/Async-std runtime for async operations?

No, I just haven't looked into it yet (and there's no async drop so we're a bit limited).

PR #156, opened 2021-11-20
Replace unsafe code from util

ironhaven:
This patch does two things

1. Reverts #141 by removing hand spun thread local SmallRng that used unsafe
   P.S. Not sure but is it a security vulnerability to use a non CSPRNG for temporary filenames?
2. Replace use of str::from_utf8_unchecked with char::encode_utf8
   Examining the assembly shows no panics generated from encode_utf8

The only functional difference of this patch is replacing SmallRng with
a marginally slower RNG

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ description = "A library for managing temporary files and directories."
 
 [dependencies]
 cfg-if = "1"
-rand = { version = "0.8", features = ["small_rng", "getrandom"], default_features = false }
+rand = "0.8"
 remove_dir_all = "0.5"
 
 [target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,5 +1,5 @@
 use rand::distributions::Alphanumeric;
-use rand::{self, Rng};
+use rand::Rng;
 use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::io;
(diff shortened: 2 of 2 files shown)


Stebalien:
1. I'm happy to take a patch to make using SmallRng opt-in, but I'd like to keep it as an option.
2. I'm hesitant to remove `str::from_utf8_unchecked`. Unsafe code isn't _bad_, it just means "be careful". Have you profiled it?

ironhaven:
Thanks for taking a look. I will admit that i have not profiled the code but diffing the assembly output show the safe version only adds two instructions and a branch that will never be called and thus will get branch predicted.
```
; al holds b in the for each loop
test al,al
js <top of the loop>
```

The main reason i wrote this code is the idea if you can prove to the compiler by writing a program in safe rust that is a universal positive unless it impacts runtime greatly. I also noticed how the thread local rng was copied directly from the rand crate that was already a dependency. 

I am going to benchmark the code to see if the unsafe code pulls it's weight.

Stebalien:
> The main reason i wrote this code is the idea if you can prove to the compiler by writing a program in safe rust that is a universal positive unless it impacts runtime greatly.

It's probably not a huge issue here, so I'm fine with it. But in general, I get a bit frustrated with the "remove all unsafe code" movement as it tends to be more dogmatic than practical.

> I also noticed how the thread local rng was copied directly from the rand crate that was already a dependency.

With all the default features turned off. Try `cargo tree` on both versions to see the difference.

ironhaven:
Here is the benchmark [results and source code](https://gist.github.com/ironhaven/6b0a1dfc67257694f15a5d09c3c8ad2e) 

This is the comparison between `encode_utf8` and `from_utf8_unchecked`
violin
There not much (<2 ns) difference between the two implementations. The version that is fastest sometimes changes between runs even. The safe `u8` to `&str`code does not impact performance.

Rng comparison
violin
Using `SmallRng` is consistently 10 ns faster or put another way 10% faster than the `StdRng` considering the function only takes 100ns. Take this with a grain of salt because i benchmarked the total time to create a tempfile at about 276,000 ns.

I hope this quick benchmark shows acceptable performance number. Now you can make a informed decision if this patch is right for this codebase

Stebalien:
The change utf8 encoding looks fine, but I still want to keep an (off by default) feature for small_rng. The concern was less about performance and more about the size of the dependency tree and the compiled binary.

Stebalien:
So, after looking at both Python and Go's temporary file generators, I've decided to double down on "simple" randomness and am switching to fastrand #162.

PR #158, opened 2021-12-17
Reduce size of TempDir and TempPath

NobodyXu:
- Reduce size of `TempDir` and `TempPath` by using `Box<Path>` instead of `PathBuf`, since neither `TempDir` nor `TempPath` uses any function specific to `PathBuf` rather than deallocating them.
   
   For `TempDir::into_path` and `TempPath::into_path`, it is converted back into `PathBuf` with zero cost.
 - Reduce size of `TempDir` by removing use of `Option` and uses `mem::forget` and `mem::ManuallyDrop` to avoid calling `Drop::drop`, and use set field `TempDir::path` to an empty `Box`: `PathBuf::new().into_boxed_path()` to deallocate the heap memory.

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -9,6 +9,7 @@
 // except according to those terms.
 
 use remove_dir_all::remove_dir_all;
+use std::mem;
 use std::path::{self, Path, PathBuf};
 use std::{fmt, fs, io};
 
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
LGTM. `into_boxed_path` could allocate in some cases (I'm pretty sure it shrinks the underlying buffer) but that's not the end of the world.

Stebalien:
Thanks!

PR #159, opened 2021-12-21
Add method for manually creating TempPaths

exactly-one-kas:


src/file/mod.rs
```diff
@@ -351,6 +351,18 @@ impl TempPath {
...
         }
     }
+
+    /// Create a new TempPath from an existing path. This can be done even if no
+    /// file exists at the given path.
+    ///
+    /// This is mostly useful for interacting with libraries and external
+    /// components that provide files to be consumed or expect a path with no
+    /// existing file to be given.
+    pub fn from_path(path: impl AsRef<Path>) -> Self {
```
Stebalien, line 361: nit: This should probably be `pub fn from_path(path: impl Into<PathBuf>)`. That should take most of the same types (really, anything that implements `AsRef<OsStr>`, but it will allow us to take ownership without allocating.

Stebalien (approved):
One nit but this seems like a useful feature otherwise.

src/file/mod.rs
Stebalien: Usually I'd just take an `AsRef<Path>` to be consistent across the API, but this _is_ a `from`-like method.

exactly-one-kas: Fair point

Stebalien (approved):

Stebalien:
Thanks!

PR #161, opened 2022-01-06
Expose `SpooledInner` and add a `into_inner()`.

ryanmcgrath:
Heya! Really appreciate this library - using it for something and wound up with a case where I needed to extract the underlying storage on a `SpooledTempFile`.

In my case, once I've finished writing it, I need to pass it off to an async runtime to work with. Rather than deal with implementing `AsyncRead`/`AsyncWrite` on `SpooledTempFile`, I found it easier to simply pull out the underlying type and use the general `From`/`Into` conversion methods to get to an async file/cursor.

This effectively sidesteps dragging a bunch of async-related crates into this while still enabling a path towards bridging them. Something you'd be open to merging?

### DIFF
--- a/src/spooled.rs
+++ b/src/spooled.rs
@@ -4,7 +4,7 @@ use std::io::{self, Cursor, Read, Seek, SeekFrom, Write};
 
 /// A wrapper for the two states of a `SpooledTempFile`.
 #[derive(Debug)]
-pub enum SpooledInner {
+pub enum SpooledData {
     InMemory(Cursor<Vec<u8>>),
     OnDisk(File),
 }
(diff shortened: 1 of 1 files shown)


Stebalien:
If you can come up with a better name than `SpooledInner`, this seems reasonable.

However, I'd actually like async support as well (behind feature flags).

ryanmcgrath:
Heh, async support would be awesome - just outside of the scope of what I can contribute right now. :(

How's one of the following sound? Ordered best idea to worst, I guess.

- `SpooledStorage`
- `SpooledData`
- `SpooledIO`
- `SpooledReaderWriter`

Stebalien:
`SpooledData` seems reasonable. Not entirely accurate, but close enough.

ryanmcgrath:
Updated. Thanks again!

Stebalien (approved):

Stebalien:
No, thank you!

PR #162, opened 2022-01-08
replace rand with fastrand

Stebalien:
Re-opens #118.

At the time the PR was submitted, fastrand was new. But it's been a year now since we switched to
SmallRng and it's still going strong.

While this will _slightly_ increase the build size for crates already depending on rand, fastrand is
really small and has no direct dependencies.

Security: Neither Go nor Python's tempfile name generator use cryptographic randomness, so I'm feeling pretty comfortable sticking with non-cryptographic randomness here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ description = "A library for managing temporary files and directories."
 
 [dependencies]
 cfg-if = "1"
-rand = { version = "0.8", features = ["small_rng", "getrandom"], default_features = false }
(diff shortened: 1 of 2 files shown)


Stebalien:
FYI @joshuarli. Thanks for implementing this and sorry for making you implement the SmallRng version.

I figured I'd take advantage of the fact that I've been an absentee maintainer to reconsider this decision before releasing 3.3.0.

PR #166, opened 2022-01-11
Port to rustix, and add support for using Linux's `renameat2`

sunfishcode:
This ports tempfile to use the rustix crate instead of making direct libc calls, which factors out an `unsafe` block, simplifies error handling, and simplifies string handling.

While here, it also implements a suggestion in a comment; on recent Linux it uses `renameat2` with `RENAME_NOREPLACE` to implement `persist` in non-overwrite mode.

This does require incrementing the minimum required Rust version to 1.48. Please feel free to decline this PR if you don't wish to take on these new requirements.

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -105,9 +105,30 @@ pub fn persist(old_path: &Path, new_path: &Path, overwrite: bool) -> io::Result<
     if overwrite {
         renameat(cwd(), old_path, cwd(), new_path)?;
     } else {
+        // On Linux, use `renameat_with` to avoid overwriting an existing name,
+        // if the kernel and the filesystem support it.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        {
+            use rustix::fs::{renameat_with, RenameFlags};
+            use rustix::io::Errno;
+            use std::sync::atomic::{AtomicBool, Ordering::Relaxed};
+
+            static NOSYS: AtomicBool = AtomicBool::new(false);
+            if !NOSYS.load(Relaxed) {
+                match renameat_with(cwd(), old_path, cwd(), new_path, RenameFlags::NOREPLACE) {
+                    Ok(()) => return Ok(()),
+                    Err(Errno::NOSYS) => NOSYS.store(true, Relaxed),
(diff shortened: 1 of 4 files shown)


src/file/imp/unix.rs
```diff
@@ -124,25 +102,36 @@ pub fn reopen(_file: &File, _path: &Path) -> io::Result<File> {
...
+                    Ok(()) => return Ok(()),
+                    Err(Errno::NOSYS) => NOSYS.store(true, Relaxed),
+                    Err(Errno::INVAL) => {}
+                    Err(e) => return Err(e.into()),
+                }
+            }
         }
-        Ok(())
+
+        // Otherwise use `linkat` to create the new filesysten name, which
```
polarathene, line 127:
```suggestion
        // Otherwise use `linkat` to create the new filesystem name, which
```

polarathene:
@Stebalien any concerns delaying review on this PR?

---

On a related note (due to the PR author), next month Rust 1.63 should have the new I/O Safety traits in `std`.

Until it's acceptable to raise the MSRV to >=1.63 to use those traits directly from `std`, in the meantime [`io-lifetimes` could be adopted](https://github.com/sunfishcode/io-lifetimes/issues/38) to cover these sections AFAIK?:

https://github.com/Stebalien/tempfile/blob/716e58e9fc98c461c801a484dc41c59e3f6c36af/src/file/mod.rs#L968-L988

https://github.com/Stebalien/tempfile/blob/716e58e9fc98c461c801a484dc41c59e3f6c36af/src/file/imp/windows.rs#L49-L64

@sunfishcode may offer to assist with that as they've expressed interest in contributing this improvement to popular crates, but it probably helps if their contribution efforts aren't left to stagnate without feedback? :sweat_smile:

Stebalien (approved):
Sorry, this fell off my radar. LGTM pending CI (which I'm trying to get to run now...).

Stebalien:
> Until it's acceptable to raise the MSRV to >=1.63 to use those traits directly from std, in the meantime https://github.com/sunfishcode/io-lifetimes/issues/38 to cover these sections AFAIK?:

This seems reasonable to me. Just make sure to ping me when it's ready for review.

PR #167, opened 2022-01-14
Implements Clone for TempDir

hwsimmons17:
I needed this for a project we are working on. Submitting as a PR so everyone can enjoy the benefits of being able to clone a TempDir


Stebalien:
That would cause it to be deleted multiple times on drop. If you need to share it, you need to wrap it in an `Rc` or an `Arc`.

PR #168, opened 2022-02-01
Add perission documentation to NamedTempFile

MarinPostma:
I added some documentation to `NamedTempFile` regarding the permission it is created with on unox systems.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -543,6 +543,10 @@ impl NamedTempFile {
     /// platforms so please read the security section of this type's
     /// documentation.
     ///
+    /// On Unix systems, the file is created with `600` permissions, instead of the default `644`,
+    /// so that only the current user can read an write to it.
+    /// These permission remain if the file is persisted later on.
+    ///
     /// Reasons to use this method:
     ///
     ///   1. The file has a short lifetime and your temporary file cleaner is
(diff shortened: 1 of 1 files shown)


src/file/mod.rs
```diff
@@ -686,6 +690,9 @@ impl NamedTempFile {
     /// secure in the in all cases. Please read the security section on the top
     /// level documentation of this type for details.
     ///
+    /// When calling this method on Unix systems, the file keeps the `600` permissions
```
Stebalien, line 693: I'd add a new section for this, right above security.

Also, while technically correct, this statement is misleading. We avoid setting permissions because we don't really know _what_ to do:

1. If we set them _first_, we could end up setting permissive permissions in some globally accessible directory.
2. If we set them afterwards, the file will temporarily have the wrong permissions.
3. We don't actually _know_ what permissions the user wants.

I think it's safest to just state that the file's permissions will be _preserved_ and that the temporary file's default permissions are 0600.

src/file/mod.rs
```diff
@@ -543,6 +543,10 @@ impl NamedTempFile {
     /// platforms so please read the security section of this type's
     /// documentation.
     ///
+    /// On Unix systems, the file is created with `600` permissions, instead of the default `644`,
```
Stebalien, line 546: - Let's put this in a new "permissions" section.
- The default isn't necessarily 0644, it depends on the user's mode.
- The leading 0 is important (it means that the number is on octal).

Stebalien (commented):
Thanks!

PR #170, opened 2022-02-02
feat: Add TempDir::new_at

Nukesor:
This PR adds the possibility to initialize temporary directories at a very specific path without any randomization.

While writing tests for it, I decided to use your `tmpname` function to prevent any namespace clashes during testing.
However, to do so, I had to publicly expose it. To prevent any usage by your users, I simply hid it in the documentation.

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -264,6 +264,40 @@ impl TempDir {
         Builder::new().tempdir_in(dir)
     }
 
+    /// Attempts to make a temporary directory at the specified `path`.
+    /// The directory and everything inside it will be automatically
+    /// deleted once the returned `TempDir` is destroyed.
+    ///
+    /// # Errors
+    ///
+    /// If the directory can not be created, `Err` is returned.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::fs::{self, File};
+    /// use std::io::Write;
+    /// use tempfile::TempDir;
+    ///
+    /// # use std::io;
+    /// # fn run() -> Result<(), io::Error> {
+    /// // Create a directory inside of the current directory
+    /// let tmp_dir = TempDir::new_at("./temp_dir")?;
+    /// let file_path = tmp_dir.path().join("my-temporary-note.txt");
+    /// let mut tmp_file = File::create(file_path)?;
+    /// writeln!(tmp_file, "Brian was here. Briefly.")?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn new_at<P: AsRef<Path>>(path: P) -> io::Result<TempDir> {
+        let path = path.as_ref().to_owned();
+        fs::create_dir(&path)
+            .with_err_path(|| &path)
+            .map(|_| TempDir {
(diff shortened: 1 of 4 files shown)


src/dir.rs
```diff
@@ -264,6 +264,40 @@ impl TempDir {
...
+    ///
+    /// ```
+    /// use std::fs::{self, File};
+    /// use std::io::Write;
+    /// use tempfile::TempDir;
+    ///
+    /// # use std::io;
+    /// # fn run() -> Result<(), io::Error> {
+    /// // Create a directory inside of the current directory
+    /// let tmp_dir = TempDir::new_at("./temp_dir")?;
```
Stebalien, line 285: This will modify the current directory when running the tests. Let's put this in `std::env::temp_dir`.

src/util.rs
```diff
@@ -5,7 +5,9 @@ use std::{io, iter::repeat_with};
 
 use crate::error::IoResultExt;
 
-fn tmpname(prefix: &OsStr, suffix: &OsStr, rand_len: usize) -> OsString {
+// Hide this function in the docs as it's only exposed for usage in integration tests.
+#[doc(hidden)]
+pub fn tmpname(prefix: &OsStr, suffix: &OsStr, rand_len: usize) -> OsString {
```
Stebalien, line 10: Instead of exposing this function, I'd just create a temporary directory inside a temporary directory. I.e.:

1. `let tmp = TempDir:new()`.
2. `TempDir::new_at(tmp.path().join("my_name")`

src/dir.rs
```diff
@@ -264,6 +264,40 @@ impl TempDir {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn new_at<P: AsRef<Path>>(path: P) -> io::Result<TempDir> {
+        let path = path.as_ref().to_owned();
+        fs::create_dir(&path)
+            .with_err_path(|| &path)
+            .map(|_| TempDir {
+                path: path.into_boxed_path(),
+            })
```
Stebalien, line 298: use the existing `create` method (in this file).

src/dir.rs
```diff
@@ -264,6 +264,40 @@ impl TempDir {
...
+    /// # fn run() -> Result<(), io::Error> {
+    /// // Create a directory inside of the current directory
+    /// let tmp_dir = TempDir::new_at("./temp_dir")?;
+    /// let file_path = tmp_dir.path().join("my-temporary-note.txt");
+    /// let mut tmp_file = File::create(file_path)?;
+    /// writeln!(tmp_file, "Brian was here. Briefly.")?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn new_at<P: AsRef<Path>>(path: P) -> io::Result<TempDir> {
```
Stebalien, line 292: There should be a `tempdir_at` function on the builder (like `tempdir_in`), and this method should call that method.

`Builder::tempdir_at` will also need to resolve the temporary directory like `tempdir_in` does (in case the user changes temporary directories).

Stebalien (changes requested):

PR #171, opened 2022-02-02
feat: flag to disable dir remove on drop

### LINKED ISSUE #169: [Feature] Flag on tempdir to not delete the directory on `Drop`
`TempDir` is often used in tests to get an isolated environment.

When tests fail, it's sometimes extremely helpful to look at the failed test's artifacts.
However, this would have to be achieved without necessarily aborting the test (via `Err` or `panic`).

For instance, it would be quite helpful to disable the `drop` if some suspicious, but not outright wrong, values are detected.
This would allow the test to finish as usual, but one can still take a look at what might have caused those values afterwards.

To easily achieve this, it would be nice to have some kind of flag to disable the dir removal functionality.

Nukesor:
Resolves #169

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -194,6 +194,7 @@ pub fn tempdir_in<P: AsRef<Path>>(dir: P) -> io::Result<TempDir> {
 /// [`std::process::exit()`]: http://doc.rust-lang.org/std/process/fn.exit.html
 pub struct TempDir {
     path: Box<Path>,
+    drop: bool,
 }
 
 impl TempDir {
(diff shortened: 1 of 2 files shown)


Stebalien:
Can you not just call `into_path()`?

Dushistov:
> Can you not just call into_path()?

This would require extra step to delete temporary directory.
Normal scenario:
```
let tmp = tempdir().unwrap();
tmp.set_delete_on_drop(false);
println!("Path if test failed: {}", tmp.path().display());
test_code();
tmp.set_delete_on_drop(true);
tmp.close().unwrap();
```
So if all ok temporary directory was deleted and if something failed, we have files to 
dig into problem.
So if exists "counter part" like `impl From<&Path> for TempDir` then `into_path` would be enough.

Nukesor:
The original described scenario also was, that the test should continue as expected in case of an error case.

```rust
let tmp = tempdir().unwrap();
let somethings_weird = do_some_thing(&tmp);

// Something's weird. Keep the files for later inspection but let the code run as usual.
if somethings_weird {
    tmp.set_drop(false);
    log::warning("Something might be off. Investigate the files at path: {:?}", tmp.path());
}

do_some_other_thing(&tmp);

Ok(())
```

Calling `.into_path()` wouldn't work in this case.

Stebalien:
This seems quite specific to your use-case and, unfortunately, increases the size/complexity of the tempdir object.

I suggest you create a wrapper type with this feature or, even better, wrap your tests in something that only removes the temporary file if some atomic "failed" flag hasn't been set. That'll give you the precise control you need.

Stebalien:
To give you a bit of context, I'm trying to balance features against complexity. Functions like `new_at` are easy to accept because they don't really add any new complexity or change how anything works under the covers. A feature like this, on the other hand, would need motivation both in the form of "it's generally useful" and "there's no reasonable way to do this externally".

PR #172, opened 2022-03-29
Features

rbtcollins:
Fix compile warnings on Windows, compile failures in 'cargo build --all-features --all-targets' and fmt Cargo.toml.


Stebalien (approved):
Thanks for the fixes!

I've reverted the TOML change as it conflicts with other changes (and the current ordering matches most rust projects that put the name first).

PR #173, opened 2022-03-29
Fix clippy & markdown lints

rbtcollins:


tests/tempfile.rs
```diff
@@ -2,8 +2,11 @@
 
 use std::fs;
 use std::io::{Read, Seek, SeekFrom, Write};
-use std::sync::mpsc::{sync_channel, TryRecvError};
-use std::thread;
+#[cfg(target_os = "linux")]
+use std::{
```
Stebalien, line 6: I'd just move these into the only test that needs them.

PR #177, opened 2022-04-08
Add NamedTempFile<F>, Builder::{make,make_in}, and NamedTempFile::from_parts

### LINKED ISSUE #175: Support for temporary UNIX domain sockets
I have a case where I want to create a temporary socket file which should be deleted when the server shuts down. `tempfile::NamedTempFile` doesn't work for this because you can't call [`UnixListener::bind(path)`](https://doc.rust-lang.org/std/os/unix/net/struct.UnixListener.html#method.bind) on a file that already exists. It'd be great to have `tempfile` handle the file path construction, but delegate the `bind` to a callback function.

For example, the API could look something like this:
```rust
impl Builder {
    // The contract is that `F` must return `io::ErrorKind::AlreadyExists` or
    // `io::ErrorKind::AddrInUse` (in the case of sockets) to continue retrying.
    pub fn make<F, R>(&self, f: F) -> io::Result<R>
    where
        F: Fn(PathBuf) -> io::Result<R>,
    {
        self.make_in(&env::temp_dir(), f)
    }

    pub fn make_in<F, R, P>(&self, dir: P, f: F) -> io::Result<R>
    where
        F: Fn(PathBuf) -> io::Result<R>,
        P: AsRef<Path>,
    {
        util::create_helper(/* ... */)
    }
}
```

And used like this:
```rust
    let (listener, path) = tempfile::Builder::new()
        .make(|path| Ok((UnixListener::bind(&path)?, TempPath::from_path(path))))?;
```

If `NamedTempFile` gained a generic parameter (e.g., `NamedTempFile<F = File>`), then we could put the socket inside of that. I *think* this generic parameter could be a backward-compatible change, but I'm not sure. Such a change could come later, since the tuple above should work fine.
```rust
    let listener = tempfile::Builder::new().make(|path| {
        Ok(NamedTempFile::from_parts(
            UnixListener::bind(&path)?,
            TempPath::from_path(path),
        ))
    })?;
```

Let me know what you think of this idea! This is generic enough to support more use cases beyond UNIX domain sockets, but I'm not sure what they are yet. There also shouldn't be any breaking changes introduced by this.

I'm happy to implement this if it seems reasonable.

jasonwhite:
There are two main parts to this PR. Unfortunately GitHub doesn't really have the ability to make PRs depend on each other. Let me know if you want this split up into separate PRs.

Fixes #175

## `NamedTempFile<F = File>`

This adds a generic parameter to `NamedTempFile`. This enables using it with non-File types, such as:
 - `std::os::unix::net::UnixListener`,
 - `tokio::net::UnixListener`, or
 - `tokio::fs::File`

This should be a backwards compatible change. To do this, some `File`-specific stuff is only implemented for `NamedTempFile<File>`.

## `Builder::{make,make_in}`

This enables the user to leverage the file-path building part of the library while providing their own custom file creation callback. This is useful for creating UNIX domain sockets.

It was mentioned in #175 to provide a helper for creating sockets, but since UNIX domain sockets are only on UNIX, I believe it would be the only platform-dependent function. Thus, I held off on making that change. Happy to reconsider this though.

Note that `NamedTempFile::from_parts` was added to implement `Builder::make_in`. This could be made a private function, but it seems useful in the public API as well.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -478,18 +478,18 @@ impl AsRef<OsStr> for TempPath {
 /// [`NamedTempFile::new_in()`]: #method.new_in
 /// [`std::env::temp_dir()`]: https://doc.rust-lang.org/std/env/fn.temp_dir.html
 /// [`std::process::exit()`]: http://doc.rust-lang.org/std/process/fn.exit.html
-pub struct NamedTempFile {
+pub struct NamedTempFile<F = File> {
     path: TempPath,
-    file: File,
+    file: F,
 }
 
-impl fmt::Debug for NamedTempFile {
+impl<F> fmt::Debug for NamedTempFile<F> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "NamedTempFile({:?})", self.path)
     }
 }
 
-impl AsRef<Path> for NamedTempFile {
+impl<F> AsRef<Path> for NamedTempFile<F> {
     #[inline]
     fn as_ref(&self) -> &Path {
         self.path()
@@ -497,41 +497,46 @@ impl AsRef<Path> for NamedTempFile {
 }
 
 /// Error returned when persisting a temporary file fails.
-#[derive(Debug)]
-pub struct PersistError {
+pub struct PersistError<F = File> {
     /// The underlying IO error.
     pub error: io::Error,
     /// The temporary file that couldn't be persisted.
-    pub file: NamedTempFile,
+    pub file: NamedTempFile<F>,
 }
 
-impl From<PersistError> for io::Error {
+impl<F> fmt::Debug for PersistError<F> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "PersistError({:?})", self.error)
+    }
+}
+
+impl<F> From<PersistError<F>> for io::Error {
     #[inline]
-    fn from(error: PersistError) -> io::Error {
+    fn from(error: PersistError<F>) -> io::Error {
         error.error
     }
 }
 
-impl From<PersistError> for NamedTempFile {
+impl<F> From<PersistError<F>> for NamedTempFile<F> {
     #[inline]
-    fn from(error: PersistError) -> NamedTempFile {
+    fn from(error: PersistError<F>) -> NamedTempFile<F> {
         error.file
     }
 }
 
-impl fmt::Display for PersistError {
+impl<F> fmt::Display for PersistError<F> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "failed to persist temporary file: {}", self.error)
     }
 }
 
-impl error::Error for PersistError {
+impl<F> error::Error for PersistError<F> {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         Some(&self.error)
     }
 }
 
-impl NamedTempFile {
+impl NamedTempFile<File> {
     /// Create a new named temporary file.
     ///
     /// See [`Builder`] for more configuration.
@@ -601,7 +606,9 @@ impl NamedTempFile {
     pub fn new_in<P: AsRef<Path>>(dir: P) -> io::Result<NamedTempFile> {
         Builder::new().tempfile_in(dir)
     }
+}
 
+impl<F> NamedTempFile<F> {
     /// Get the temporary file's path.
     ///
     /// # Security
@@ -711,7 +718,7 @@ impl NamedTempFile {
     /// ```
     ///
     /// [`PersistError`]: struct.PersistError.html
-    pub fn persist<P: AsRef<Path>>(self, new_path: P) -> Result<File, PersistError> {
+    pub fn persist<P: AsRef<Path>>(self, new_path: P) -> Result<F, PersistError<F>> {
         let NamedTempFile { path, file } = self;
         match path.persist(new_path) {
             Ok(_) => Ok(file),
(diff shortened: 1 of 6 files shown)


src/file/mod.rs
```diff
@@ -858,54 +895,21 @@ impl NamedTempFile {
...
 
-impl Read for NamedTempFile {
+impl<F: Read> Read for NamedTempFile<F> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         self.as_file_mut().read(buf).with_err_path(|| self.path())
     }
 }
 
-impl<'a> Read for &'a NamedTempFile {
+impl<'a> Read for &'a NamedTempFile<File> {
```
jasonwhite, line 914: This is a weird one. I couldn't figure out how to implement this generically. Normally, you can't do mutation with a immutable reference. The `File` type implements it at: https://doc.rust-lang.org/std/fs/struct.File.html#impl-Read-1

Same is true for `impl Write for &'a File`.

tests/namedtempfile.rs
```diff
@@ -326,3 +338,86 @@ fn test_keep() {
...
+#[cfg(unix)]
+#[test]
+fn test_make_uds() {
+    use std::os::unix::net::UnixListener;
+
+    let temp_sock = Builder::new()
+        .prefix("tmp")
+        .suffix(".sock")
+        .rand_bytes(12)
+        .make(|path| UnixListener::bind(path))
```
jasonwhite, line 372: Note: I couldn't get `.make(UnixListener::bind)` to work because of the following error:
```
error: implementation of `FnOnce` is not general enough
   --> tests/namedtempfile.rs:372:10
    |
372 |         .make(UnixListener::bind)
    |          ^^^^ implementation of `FnOnce` is not general enough
    |
    = note: `fn(&'2 Path) -> Result<UnixListener, std::io::Error> {UnixListener::bind::<&'2 Path>}` must implement `FnOnce<(&'1 Path,)>`, for any lifetime `'1`...
    = note: ...but it actually implements `FnOnce<(&'2 Path,)>`, for some specific lifetime `'2`
```

src/file/mod.rs
```diff
@@ -498,40 +498,40 @@ impl AsRef<Path> for NamedTempFile {
...
 
-impl fmt::Display for PersistError {
+impl<F> fmt::Display for PersistError<F> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "failed to persist temporary file: {}", self.error)
     }
 }
 
-impl error::Error for PersistError {
+impl<F: fmt::Debug> error::Error for PersistError<F> {
```
Stebalien, line 528: Is there any reason the underlying file type needs to be "debug"?

src/lib.rs
```diff
@@ -534,4 +534,123 @@ impl<'a, 'b> Builder<'a, 'b> {
 
         util::create_helper(dir, self.prefix, self.suffix, self.random_len, dir::create)
     }
+
+    /// Attempts to create a temporary file (or file-like object) using the
+    /// provided closure inside of [`std::env::temp_dir()`]. The closure is
```
Stebalien, line 539:
```suggestion
    /// provided closure inside of creating a [`std::fs::File`]. The closure is
```

src/lib.rs
```diff
@@ -534,4 +534,123 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+    /// This can be helpful for taking full control over the file creation, but
+    /// leaving the temporary file path construction up to the library. This
+    /// also enables creating a temporary UNIX domain socket, since it is not
+    /// possible to bind to a socket that already exists.
+    ///
+    /// Note that [`Builder::append`] is ignored when using [`Builder::make`].
+    ///
+    /// # Security
+    ///
+    /// See [the security][security] docs on `NamedTempFile`.
```
Stebalien, line 555: Also note that this function will not:

- Ensure a safe mode.
- That the file doesn't already exist.

Both of these are responsibilities of closure.

src/lib.rs
```diff
@@ -534,4 +534,123 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+    /// let tempsock = Builder::new().make_in("./", |path| UnixListener::bind(path))?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// [security]: struct.NamedTempFile.html#security
+    /// [resource-leaking]: struct.NamedTempFile.html#resource-leaking
+    pub fn make_in<F, R, P>(&self, dir: P, f: F) -> io::Result<NamedTempFile<R>>
+    where
+        F: Fn(&Path) -> io::Result<R>,
```
Stebalien, line 640: `FnMut` is more general and may help users in some cases.

tests/namedtempfile.rs
Stebalien: Hm. Oh well.

src/lib.rs
Stebalien: And below.

Stebalien (approved):
Some small nits, otherwise LGTM.

src/file/mod.rs
Stebalien: Committed a fix. The trick is `where &F: Read`.

jasonwhite: Awesome! Thanks for that fix.

src/file/mod.rs
jasonwhite: It's because of the `#[derive(Debug)]`. I'll do a manual impl of `Debug` instead to avoid this.

src/lib.rs
jasonwhite: That wording still looks a little odd (`s/inside/instead/`?). I'll try to make this clearer.

Stebalien: Uh, yes.

Stebalien: Oh, wait, nvm. I just read this as "instead of". No, the current wording makes sense. Although re-ordering to it as "temporary file inside of ... with the provided closure" may be easier (at least for me).

src/lib.rs
jasonwhite: Good call. Fixed that and added a test for it.

src/lib.rs
jasonwhite: Good call. I added a bunch of docs for the safety considerations and responsibilities of the caller.

src/lib.rs
jasonwhite: I reworded this. It should be clearer now.

jasonwhite:
Thanks for the review! All review comments should be addressed. Let me know if there's anything else.

Stebalien (approved):

PR #180, opened 2022-06-10
Add `#[must_use]` to various functions

joshtriplett:
Based on encountering some code that called `TempDir::path` and ignored
the result.


Stebalien (approved):
Hm. It would be _nice_ if return values were "must use" by default with the ability to add an `#[optional_return]` attribute...

Issue #183, opened 2022-07-12
doc: lazy_static NamedTempFile fails to remove itself because destructor doesn't run


jtmoon79:
A `static ref NamedTempFile` initialized using [`lazy_static`](https://github.com/rust-lang-nursery/lazy-static.rs) fails to remove itself after program completion.

### Reproduction

1. note the contents of the temporary directory (e.g. `ls -ltr /tmp`)
2. run the following [Rust Code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82b9329aa060b9df7abfc0b628a1f60e)
```rust
extern crate tempfile; // 3.3.0
use tempfile::NamedTempFile;

extern crate lazy_static; // 1.4.0
use lazy_static::lazy_static;

use std::io::Write;  // for `NamedTempFile.write_all`


lazy_static! {
    static ref NTF1: NamedTempFile = {
        let mut ntf = NamedTempFile::new().unwrap();
        ntf.write_all("foo".as_bytes());
        ntf
    };
}

fn main() {
    eprintln!("NTF1.path() {:?}", NTF1.path());
}
```
3. note the contents of the temporary directory (e.g. `ls -ltr /tmp`)
**_The temporary file printed by the program remains after program completion._**

Stebalien:
This is known and, unfortunately, unfixable as destructors don't run for statics. See https://docs.rs/tempfile/latest/tempfile/#resource-leaking.

Improvements to the documentation are welcome.

jtmoon79:
> Improvements to the documentation are welcome.

I'm new to generating rust documentation. What is the full command(s) you use to generate the documentation?

I forked and cloned and edited some docstrings. Then ran `cargo build` which succeeded.
But `rustdoc` fails.
```shell
$ pwd
~/Projects/tempfile

$ cargo build
...

$ rustdoc --version
rustdoc 1.61.0-beta.3 (2431a974c 2022-04-17)

$ rustdoc src/lib.rs --crate-name tempfile
error[E0432]: unresolved import `remove_dir_all`
  --> src/dir.rs:11:5
   |
11 | use remove_dir_all::remove_dir_all;
   |     ^^^^^^^^^^^^^^ maybe a missing crate `remove_dir_all`?

error[E0432]: unresolved import `fastrand`
 --> src/util.rs:1:5
  |
1 | use fastrand;
  |     ^^^^^^^^ no `fastrand` in the root

[...]

PR #184, opened 2022-07-13
doc clarify statics may leak

jtmoon79:
documentation notes static declared instances may leak (like with
lazy_static). Links to underlying reason
https://github.com/rust-lang-nursery/lazy-static.rs/issues/62

Issue #183

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -467,17 +467,19 @@ impl AsRef<OsStr> for TempPath {
 /// # Resource Leaking
 ///
 /// If the program exits before the `NamedTempFile` destructor is
-/// run, such as via [`std::process::exit()`], by segfaulting, or by
-/// receiving a signal like `SIGINT`, then the temporary file
-/// will not be deleted.
+/// run then the temporary file will not be deleted. This can happen
+/// if the process exits using [`std::process::exit()`], a segfault occurs,
+/// receiving an interrupt signal like `SIGINT` that is not handled, or by using
+/// a statically declared `NamedTempFile` instance (like with [`lazy_static`]).
 ///
-/// Use the [`tempfile()`] function unless you absolutely need a named file.
+/// Use the [`tempfile()`] function unless you need a named file or a file with a path.
 ///
 /// [`tempfile()`]: fn.tempfile.html
 /// [`NamedTempFile::new()`]: #method.new
 /// [`NamedTempFile::new_in()`]: #method.new_in
 /// [`std::env::temp_dir()`]: https://doc.rust-lang.org/std/env/fn.temp_dir.html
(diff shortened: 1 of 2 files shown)


src/file/mod.rs
```diff
@@ -467,17 +467,19 @@ impl AsRef<OsStr> for TempPath {
...
-/// run, such as via [`std::process::exit()`], by segfaulting, or by
-/// receiving a signal like `SIGINT`, then the temporary file
-/// will not be deleted.
+/// run then the temporary file will not be deleted. This can happen
+/// if the process exits using [`std::process::exit()`], a segfault occurs,
+/// receiving an interrupt signal like `SIGINT` that is not handled, or by using
+/// a statically declared `NamedTempFile` instance (like with [`lazy_static`]).
 ///
-/// Use the [`tempfile()`] function unless you absolutely need a named file.
+/// Use the [`tempfile()`] function unless you need a named file or a file with a path.
```
Stebalien, line 475: "file with a path" is the same as "named file". Feel free to drop the "named file" part if you feel that "a file with a path" is clearer.

src/file/mod.rs
```diff
@@ -467,17 +467,19 @@ impl AsRef<OsStr> for TempPath {
 /// # Resource Leaking
 ///
 /// If the program exits before the `NamedTempFile` destructor is
-/// run, such as via [`std::process::exit()`], by segfaulting, or by
-/// receiving a signal like `SIGINT`, then the temporary file
-/// will not be deleted.
+/// run then the temporary file will not be deleted. This can happen
```
Stebalien, line 470:
```suggestion
/// run, the temporary file will not be deleted. This can happen
```

(Reads better? There may be some English rule here, that wasn't my best class)

Stebalien (commented):
nits, otherwise LGTM. Thanks!

src/file/mod.rs
jtmoon79: Fixed.

src/file/mod.rs
jtmoon79: English is complicated: https://www.dailywritingtips.com/comma-before-but/

Since the implied subject is "the temporary file" then a single "then" may be used without a comma.
I'm about 70% sure I'm correct on this.

jtmoon79:
All yours. You may want to squash the PR into the first commit.

src/file/mod.rs
Stebalien: It's a separate clause.

https://www.grammarly.com/blog/conditional-sentences/

The "then" is optional and usually omitted, but the comma isn't.

jtmoon79: Fixed.

Stebalien:
Thanks! Sorry for being so picky on the wording and thanks for your patience.

PR #187, opened 2022-07-29
chore: update changelog

Stebalien:


joshtriplett:
I would love to make use of tempfile 3.4 (and thank you for working on it!). I have some changes in other crates that depend on the new APIs it exposes.

Is there anything blocking the 3.4 release? Anything I could help with?

PR #195, opened 2022-10-20
add keep support to TempDir

### LINKED ISSUE #194: feature request: add support for keeping TempDir
I would like to be able to conditionally avoid cleaning up a TempDir, and right now there doesn't seem to be an option for that besides resorting to shenanigans to avoid dropping the struct, which would likely lead to some memory leaks. It would be nice to add a `disown()` method that lets you mark a TempDir or NamedTempFile as not needing to be cleaned up on distruction.

My use case is that I'm using a TempDir as a place to throw some logs during tests and I want to be able to inspect those logs after the fact if I'm iterating on the test by setting an environment variable or something.

ethanpailes:
This patch adds keep() support to TempDir. Rather than using mem::forget() tricks like NamedTempFile does, it just uses a flag, which I think is a little easier to follow, though does require an extra byte of memory footprint.

Closes #194

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -194,6 +194,7 @@ pub fn tempdir_in<P: AsRef<Path>>(dir: P) -> io::Result<TempDir> {
 /// [`std::process::exit()`]: http://doc.rust-lang.org/std/process/fn.exit.html
 pub struct TempDir {
     path: Box<Path>,
+    disowned: bool,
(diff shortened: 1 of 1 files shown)


Stebalien:
See https://github.com/Stebalien/tempfile/issues/194#issuecomment-1286101601.

PR #197, opened 2022-10-21
add keep support to TempDir

### LINKED ISSUE #194: feature request: add support for keeping TempDir

ethanpailes:
This patch adds a keep() method to TempDir so that it better matches the interface of NamedTempFile.

Closes #194

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -334,6 +334,38 @@ impl TempDir {
         mem::replace(&mut this.path, PathBuf::new().into_boxed_path()).into()
     }
 
+    /// Marks this directory so that it will not get cleaned up when the rust
+    /// object is dropped.
+    ///
+    /// # Examples
+    /// ```
+    /// use std::fs;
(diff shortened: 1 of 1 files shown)


jayvdb:
as this is just an alias for `into_path` , perhaps all that is needed is improved docs for `into_path` to help people recognise its effect?

ethanpailes:
Yeah, I think it is best not to duplicate functionality. I should have closed this a while ago.

PR #207, opened 2023-02-24
Upgrade remove_dir_all to 0.8

Xaeroxe:
This removes the `winapi` dependency from `remove_dir_all`


mati865:
Is `remove_dir_all` still necessary?
https://github.com/Stebalien/tempfile/issues/129#issuecomment-1304452055

Xaeroxe:
Probably not no

PR #209, opened 2023-02-25
fix: clippy errors

Stebalien:


ControlCplusControlV:
Are you going to publish a new version without `remove_dir`  after this is merged?

Stebalien:
> Are you going to publish a new version without remove_dir after this is merged?

Yes.

PR #218, opened 2023-03-01
Fix and expand CI

LingMan:
- Downgrade to XCode 13 if the test runs with Rust 1.48 on macOS. Fixes the failing CI on that configuration.
 - Add rustfmt and clippy to CI
 - Add stable Rust to CI
 - Two tiny style nits.


Stebalien (approved):

LingMan:
There. The fix had the right idea, but accessing a non-existing variable isn't conductive to a passing CI :laughing:

PR #225, opened 2023-03-24
NamedTempFile: Fix infinite recursion in trait bounds

### LINKED ISSUE #224: The generic version of NamedTempFile can break compilation
The more generic version of `NamedTempFile` introduced in #177 by @jasonwhite can break compilation in some cases without even being involved itself. It seems to happen in some cases with generic where it will turn into a infinitely recursive type.

Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=671872049ef51f20f8bcc976177c6b62

if you comment the line `use tempfile;` the code will compile as expected.

edit: Made the MRE smaller.

jasonwhite:
Due to bug rust-lang/rust#96634, these generic bounds may cause an infinite recursion in the compiler, even in unrelated code.

This specializes the `Write for &NamedTempFile<F>` impl on `File` instead. This keeps the API the same as before the generic parameter `F` was added in #177, so it shouldn't break any code. If/when the compiler bug is fixed, we can switch it back to the more generic version. Very few types implement `Write for &T`, however. The only other one I know of is `UnixStream`, which probably wouldn't be used with `NamedTempFile` anyway (it's counterpart `UnixListener` is used instead).

Fixes #224.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -918,10 +918,7 @@ impl<F: Read> Read for NamedTempFile<F> {
     }
 }
 
-impl<'a, F> Read for &'a NamedTempFile<F>
-where
-    &'a F: Read,
-{
+impl Read for &NamedTempFile<File> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         self.as_file().read(buf).with_err_path(|| self.path())
     }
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
Annoying but yeah, this should be reasonable breakage.

Stebalien:
Thanks!

PR #236, opened 2023-04-24
Remove comment on SetFileInformationByHandle

### LINKED ISSUE #234: Would `SetFileInformationByHandle` even be helpful for `persist`?
In the `persist()` implementation for Windows, I noticed a comment:

```rust
// TODO: We should probably do this in one-shot using SetFileInformationByHandle but the API is
// really painful.
```

I was interested in seeing if I could work through the painful API and file a PR to implement this. However, [`SetFileInformationByHandle`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileinformationbyhandle) does not appear to allow moving a file and modifying its attributes at the same time. The `FileAttributes` are only contained in [`FILE_BASIC_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_basic_info), the destination path is only contained in [`FILE_RENAME_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_rename_info), and `SetFileInformationByHandle` only allows one of these to be passed at a time. So it would still end up taking 3 calls to `SetFileInformationByHandle`; the only benefit is that we could use the handle directly instead of going through the path. ([`NtSetInformationFile`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntsetinformationfile) is more powerful, but it can still only change one of the `FILE_BASIC_INFORMATION` or `FILE_RENAME_INFORMATION` at a time.)

As far as I can tell, the closest thing to being able to perform both operations "in one-shot" would be to use the Transactional NTFS functions, which, as the name suggests, only work with NTFS.

LegionMammal978:
The function cannot actually be used to move a file and set its attributes in one operation.

Fixes #234.

### DIFF
--- a/src/file/imp/windows.rs
+++ b/src/file/imp/windows.rs
@@ -75,9 +75,6 @@ pub fn keep(path: &Path) -> io::Result<()> {
 }
 
 pub fn persist(old_path: &Path, new_path: &Path, overwrite: bool) -> io::Result<()> {
-    // TODO: We should probably do this in one-shot using SetFileInformationByHandle but the API is
-    // really painful.
-
     unsafe {
         let old_path_w = to_utf16(old_path);
         let new_path_w = to_utf16(new_path);


Stebalien (approved):
Thanks!

PR #237, opened 2023-05-12
fix(docs): Change `dir` and `file` variable names ...

LeoniePhiline:
# fix(docs): Change `dir` and `file` variable names in `tempdir` and `tempdir_in` documentation 

to match existing comments and rest of module

## Problem statement

The documentation for [tempdir](https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html) and [tempdir_in](https://docs.rs/tempfile/latest/tempfile/fn.tempdir_in.html) use variables `dir` and `file` but refer to them as `tmp_dir` and `tmp_file` in code comments.

## Discussion

The comments could be changed to refer to `dir` and `file`. However, the rest of the module uses `tmp_dir` and `tmp_file`.

## PR changes

The variable names were changed. 

This way, they match the comments referring to these variables as well as the rest of the module documentation.

## Tests

Existing doctests succeed.

```
test src/dir.rs - dir::tempdir (line 36) ... ok
test src/dir.rs - dir::tempdir_in (line 85) ... ok
```

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -45,16 +45,16 @@ use crate::Builder;
 /// # }
 /// # fn run() -> Result<(), io::Error> {
 /// // Create a directory inside of `std::env::temp_dir()`
-/// let dir = tempdir()?;
+/// let tmp_dir = tempdir()?;
 ///
-/// let file_path = dir.path().join("my-temporary-note.txt");
-/// let mut file = File::create(file_path)?;
-/// writeln!(file, "Brian was here. Briefly.")?;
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
I don't really see the point, but there's no reason not to.

LeoniePhiline:
It's not a big deal - it's just correcting something that's wrong.

Even if it being wrong does no further harm than to trip the occasional reader's brain for a second.

Sometimes it's the small things. ;)

Stebalien:
No, thank you. It does make for cleaner docs.

PR #238, opened 2023-05-12
Derive Serde + Standard traits + Implement `Display`

blyxyas:
The [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/about.html) recommend deriving / implementing the following traits [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits):

```
Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Display, Default
```

`Debug` is already implemented and `Display` and `Copy` aren't derivable in this case.
So this PR derives all remaining traits, and implements `Display` for TempDir.

---

It also adds a "serde" feature ([C-SERDE](https://rust-lang.github.io/api-guidelines/interoperability.html#c-serde)), if it's enabled, Serde's `Serialize` and `Deserialize` are also derived.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tempfile"
-version = "3.5.0"
+version = "3.5.1"
 authors = [
   "Steven Allen <steven@stebalien.com>",
   "The Rust Project Developers",
@@ -19,6 +19,7 @@ description = "A library for managing temporary files and directories."
 [dependencies]
 cfg-if = "1"
 fastrand = "1.6.0"
+serde = { version = "1.0.163",  optional = true, features = ["derive"] }
 
 [target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 rustix = { version = "0.37.11", features = ["fs"] }
@@ -41,3 +42,4 @@ autocfg = "1"
 
 [features]
 nightly = []
+serde = ["dep:serde"]
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -13,6 +13,9 @@ use std::mem;
 use std::path::{self, Path, PathBuf};
 use std::{fmt, fs, io};
 
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
 use crate::error::IoResultExt;
 use crate::Builder;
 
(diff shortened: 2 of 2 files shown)


Stebalien:
None of these implementations make sense.

blyxyas:
I want to implement this for a custom type for my library, and also I want my library to use the Rust API Guidelines, that's why I made this PR.
That's kinda rude tbh

Stebalien:
I apologize for being short. PRs like this can get repetitive and frustrating, but I understand you were just trying to help so I'll give you a bit of context

First, PR review takes time. When you submit a PR for a new feature, you're not doing the maintainer a favor, you're asking for a favor; unless they requested the feature, you likely want it more than they do. This means the onus is on you to do as much work as possible up-front to avoid wasting their time.

The core issue with this PR is that it lacks any real motivation or forethought. A motivation both gives the reviewer context and forces you to think understand/think about the feature you're implementing.

In this specific case:

- Cloning a temp dir won't work because it won't clone the underlying directory. When the first `TempDir` gets dropped, the second will now refer to a non-existent directory. If you need this to be clonable, use an `Rc`/`Arc`.
- Serde makes no sense as the serialized temporary directory reference won't keep the temporary directory alive (similar issues to cloning).
- `Display` isn't implemented for the same reason it isn't implemented on `Path` (documented in `Path::display()`).
- A derived `Eq`/`Hash` are useless because, if you have multiple `TempDir` structures referring to the same directory, you already have an issue (same as cloning). `TempDir`s should be unique.

blyxyas:
> you're not doing the maintainer a favor, you're asking for a favor

Oh I'm very sorry for taking time out of my day, thinking that adding some more traits could be helpful. Forking your repo, implementing the change and opening a Pull Request explaining the changes. Ignorant me!

I'm sorry! Thanks for the favor of saying "None of this implementations makes sense" and closing it without any further explanation until I point that out!

> avoid wasting their time.

The change is made of a one-liner and a feature in `Cargo.toml`. Omg **such a waste of time**!
Thanks for being so kind to let me ask for this favor! So that you could just be rude and close it instantly, without an explanation!

[I'll just leave the Rust Code Of Conduct here](https://www.rust-lang.org/policies/code-of-conduct).

PR #241, opened 2023-06-30
Update to rustix 0.38 and update MSRV to 1.63

sdroege:

PR #244, opened 2023-07-04
Provide implementations for I/O safety traits unconditionally

LingMan:
The first commit is a cleanup removing the build script. Until now it gated the implementation of `AsFd` / `AsRawFd` / `AsHandle` / `AsRawHandle` on the existence of these traits in `std`. They have been stabilized with Rust 1.63 which will [soon](https://github.com/Stebalien/tempfile/pull/241) be this crate's MSRV.

I'll rebase this PR once that has landed. Until then CI will obviously fail.

The second commit enables `AsFd`/`AsRawFd` implementations on wasi, since `std` provides the traits there as well.


Stebalien (approved):
Thanks!

Stebalien:
(I'll let you rebase then merge when the tests pass)

LingMan:
Ok, slight annoyance with the import paths:
`AsFd` and friends exist since 1.63, but they were located at `std::os::unix::io` until 1.66. Now they are also exported from `std::os::fd`, which is the only documented path.

PR #247, opened 2023-08-06
Use `rename`/`unlink`/`etc. instead of `renameat`/`unlinkat`/etc.

sunfishcode:
Rustix on Haiku doesn't support `CWD`, so use the non-`at` versions of functions, since tempfile was only ever passing `CWD` to them.

For `link`, switch to `std::fs::hard_link`, since the `link` function has platform-dependent behavior and std has code to deal with that. This should also help portability to some platforms which lack `linkat`.

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -14,7 +14,10 @@ use crate::util;
 use std::path::Path;
 
 #[cfg(not(target_os = "redox"))]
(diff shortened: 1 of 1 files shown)


asomers (approved):
LGTM.  Thanks for the quick fix.

Stebalien (approved):

Stebalien (approved):

PR #248, opened 2023-08-06
fix: avoid *at functions for better compatibility

### LINKED ISSUE #246: Build failure on Haiku
Tempfile 3.7.0 fails to build on Haiku:
```
> cargo check -Z build-std --target x86_64-unknown-haiku
    Checking tempfile v3.7.0 (/usr/home/somers/src/rust/tempfile)
error[E0432]: unresolved import `rustix::fs::CWD`
  --> src/file/imp/unix.rs:17:55
   |
17 | use rustix::fs::{linkat, renameat, unlinkat, AtFlags, CWD};
   |                                                       ^^^ no `CWD` in `fs`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `tempfile` (lib) due to previous error
```
v3.6.0 worked just fine.  The problem seems to be that Rustix 0.38.0 removed `CWD`.
See https://github.com/bytecodealliance/rustix/commit/23af8a8c4ca37f6fdf602a26a67d3e32a1f64e80#diff-9f7a9943fff7afdd02920889c8f84454a8968e2f8015fd1ae06742255f82fe85 .
I suggest removing the `persist` method on Haiku.  That should be sufficient to get compilation working again.

Stebalien:
As far as I can tell, `*at` syscalls are equivalent to their counterparts when they're called relative to the current working directory, so this shouldn't matter. We only _need_ the `*at` syscall for `renameat2` to pass some flags.

fixes #246

### DIFF
--- a/src/file/imp/unix.rs
+++ b/src/file/imp/unix.rs
@@ -14,7 +14,7 @@ use crate::util;
 use std::path::Path;
 
 #[cfg(not(target_os = "redox"))]
-use rustix::fs::{linkat, renameat, unlinkat, AtFlags, CWD};
+use rustix::fs::{link, rename, unlink};
 
 pub fn create_named(path: &Path, open_options: &mut OpenOptions) -> io::Result<File> {
     open_options.read(true).write(true).create_new(true);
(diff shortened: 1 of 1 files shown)


Stebalien:
@sunfishcode could I get confirmation that this should be fine?

Stebalien:
Uh, never mind. I submitted this PR before I saw #247.

PR #257, opened 2023-09-19
Correct docs about what this is equivalent to

roy-work:
The docstring for `NamedTempFile::new_in` appears to be wrong.


Stebalien (approved):
Thanks!

PR #273, opened 2024-02-04
feat: Add `Builder::permissions()` method.

Byron:
This PR allows the user to control the file mode set for newly created `NameTempfile` instances.

This is paramount for when tempfiles are used as scratch space that will be persisted and renamed into the final destination once all writes are completed. This is an effective way to prevent half-written files on disk and to assure that readers can only observe the final result of a change.

This paradigm is implemented in the `gix-lock` crate which is used by the `gix-ref` crate for editing Git references. Currently these are created with `0o600` permissions, which can cause [issues when using `sudo`](https://github.com/stacked-git/stgit/issues/413) in downstream applications.

When this PR is merged, `gix-ref` can be adjusted to allow setting the mode [just like Git does](https://github.com/git/git/blob/master/lockfile.h#L175-L181).

### Review Notes

I did my best to adapt the least amount of code while staying true to the current organisation.  Initially I named the new method `mode`, but realized that `std::fs::Permissions` exists which would allow the API to be more general, despite only one available platform for implementation.

Let me explain the commits in some detail and why it's one more than needed:

* the first commit is a failing test that adds the new `Builder` method without an implementation. CI is red to show it's not working.
* the second commits implements it to be least invasive, which violates the principle that platform dependent code should remain in `imp`. CI is green.
* the third commit implements it so that `permissions` can be passed down to the platforms, which is more invasive, yet cleaner. CI is also green.

Please let me know which way you prefer as its also a trade-off.

### DIFF
--- a/src/dir/imp/unix.rs
+++ b/src/dir/imp/unix.rs
@@ -1,0 +1,21 @@
+use crate::error::IoResultExt;
+use crate::TempDir;
+use std::io;
+use std::path::PathBuf;
+
+pub fn create(path: PathBuf, permissions: Option<&std::fs::Permissions>) -> io::Result<TempDir> {
+    let mut dir_options = std::fs::DirBuilder::new();
+    #[cfg(not(target_os = "wasi"))]
+    {
+        use std::os::unix::fs::{DirBuilderExt, PermissionsExt};
+        if let Some(p) = permissions {
+            dir_options.mode(p.mode());
+        }
+    }
+    dir_options
+        .create(&path)
+        .with_err_path(|| &path)
+        .map(|_| TempDir {
+            path: path.into_boxed_path(),
+        })
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,6 +197,7 @@ pub struct Builder<'a, 'b> {
     prefix: &'a OsStr,
     suffix: &'b OsStr,
     append: bool,
+    permissions: Option<std::fs::Permissions>,
 }
 
 impl<'a, 'b> Default for Builder<'a, 'b> {
@@ -206,6 +207,7 @@ impl<'a, 'b> Default for Builder<'a, 'b> {
             prefix: OsStr::new(".tmp"),
             suffix: OsStr::new(""),
             append: false,
+            permissions: None,
         }
     }
 }
@@ -396,6 +398,63 @@ impl<'a, 'b> Builder<'a, 'b> {
     /// ```
     pub fn append(&mut self, append: bool) -> &mut Self {
         self.append = append;
+        self
+    }
+
+    /// The permissions to create the tempfile with.
+    /// This allows to them differ from the default mode of `0o600` on Unix.
+    ///
+    /// # Security
+    ///
+    /// By default, the permissions of tempfiles on unix are set for it to be
+    /// readable and writable by the owner only, yielding the greatest amount
+    /// of security.
+    /// As this method allows to widen the permissions, security would be
+    /// reduced in such cases.
+    ///
+    /// # Platform Notes
+    /// ## Unix
+    ///
+    /// The actual permission bits set on the tempfile will be affected by the
+    /// `umask` applied by the underlying `open` syscall.
+    ///
+    /// ## Windows and others
+    ///
+    /// This setting is ignored.
+    ///
+    /// # Limitations
+    ///
+    /// Permissions for directories aren't currently set even though it would
+    /// be possible on Unix systems.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::io;
+    /// # fn main() {
+    /// #     if let Err(_) = run() {
+    /// #         ::std::process::exit(1);
+    /// #     }
+    /// # }
+    /// # fn run() -> Result<(), io::Error> {
+    /// # use tempfile::Builder;
+    /// #[cfg(unix)]
+    /// {
+    ///     use std::os::unix::fs::PermissionsExt;
+    ///     let all_read_write = std::fs::Permissions::from_mode(0o666);
+    ///     let tempfile = Builder::new().permissions(all_read_write).tempfile()?;
+    ///     let actual_permissions = tempfile.path().metadata()?.permissions();
+    ///     assert_ne!(
+    ///         actual_permissions.mode() & !0o170000,
+    ///         0o600,
+    ///         "we get broader permissions than the default despite umask"
+    ///     );
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn permissions(&mut self, permissions: std::fs::Permissions) -> &mut Self {
+        self.permissions = Some(permissions);
         self
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -401,7 +401,7 @@ impl<'a, 'b> Builder<'a, 'b> {
         self
     }
 
-    /// The permissions to create the tempfile with.
+    /// The permissions to create the tempfile or [tempdir](Self::tempdir) with.
     /// This allows to them differ from the default mode of `0o600` on Unix.
     ///
     /// # Security
@@ -415,20 +415,18 @@ impl<'a, 'b> Builder<'a, 'b> {
     /// # Platform Notes
     /// ## Unix
     ///
-    /// The actual permission bits set on the tempfile will be affected by the
-    /// `umask` applied by the underlying `open` syscall.
+    /// The actual permission bits set on the tempfile or tempdir will be affected by the
+    /// `umask` applied by the underlying syscall.
+    ///
     ///
     /// ## Windows and others
     ///
     /// This setting is unsupported and trying to set a file or directory read-only
     /// will cause an error to be returned..
     ///
-    /// # Limitations
-    ///
-    /// Permissions for directories aren't currently set even though it would
-    /// be possible on Unix systems.
-    ///
     /// # Examples
+    ///
+    /// Create a named temporary file that is world-readable.
     ///
     /// ```
     /// # use std::io;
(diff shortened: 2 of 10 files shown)


NobodyXu:
@Byron There's a better way to create tempfile that persists after writing is done on Linux using `O_TMPFILE`.

While it only supports ext2, ext3, ext4, UDF, Minix, tmpfs 
XFS (Linux 3.15), Btrfs (Linux 3.16), F2FS (Linux 3.16), ubifs (Linux 4.9), and needs either `CAP_DAC_READ_SEARCH` or a procfs to be mounted, it's still a pretty good solution.

It prevents any other process from manipulating the file (except through procfs) until the caller process marked it as ready, and it removed the file on process exit even if the rust destructor is not called (e.g. panic during unwinding, calling `exit` directly)

```
O_TMPFILE (since Linux 3.11)
              Create an unnamed temporary regular file.  The pathname
              argument specifies a directory; an unnamed inode will be
              created in that directory's filesystem.  Anything written
              to the resulting file will be lost when the last file
              descriptor is closed, unless the file is given a name.

              O_TMPFILE must be specified with one of O_RDWR or O_WRONLY
              and, optionally, O_EXCL.  If O_EXCL is not specified, then
              linkat(2) can be used to link the temporary file into the
              filesystem, making it permanent, using code like the
              following:

                  char path[PATH_MAX];
                  fd = open("/path/to/dir", O_TMPFILE | O_RDWR,
                                          S_IRUSR | S_IWUSR);

                  /* File I/O on 'fd'... */

                  linkat(fd, "", AT_FDCWD, "/path/for/file", AT_EMPTY_PATH);

                  /* If the caller doesn't have the CAP_DAC_READ_SEARCH
                     capability (needed to use AT_EMPTY_PATH with linkat(2)),
                     and there is a proc(5) filesystem mounted, then the
                     linkat(2) call above can be replaced with:

                  snprintf(path, PATH_MAX,  "/proc/self/fd/%d", fd);
                  linkat(AT_FDCWD, path, AT_FDCWD, "/path/for/file",
                                          AT_SYMLINK_FOLLOW);
                  */

              In this case, the open() mode argument determines the file
              permission mode, as with O_CREAT.

              Specifying O_EXCL in conjunction with O_TMPFILE prevents a
              temporary file from being linked into the filesystem in
              the above manner.  (Note that the meaning of O_EXCL in
              this case is different from the meaning of O_EXCL
              otherwise.)

              There are two main use cases for O_TMPFILE:

              •  Improved tmpfile(3) functionality: race-free creation
                 of temporary files that (1) are automatically deleted
                 when closed; (2) can never be reached via any pathname;
                 (3) are not subject to symlink attacks; and (4) do not
                 require the caller to devise unique names.

              •  Creating a file that is initially invisible, which is
                 then populated with data and adjusted to have
                 appropriate filesystem attributes (fchown(2),
                 fchmod(2), fsetxattr(2), etc.)  before being atomically
                 linked into the filesystem in a fully formed state
                 (using linkat(2) as described above).

              O_TMPFILE requires support by the underlying filesystem;
              only a subset of Linux filesystems provide that support.
              In the initial implementation, support was provided in the
              ext2, ext3, ext4, UDF, Minix, and tmpfs filesystems.
              Support for other filesystems has subsequently been added
              as follows: XFS (Linux 3.15); Btrfs (Linux 3.16); F2FS
              (Linux 3.16); and ubifs (Linux 4.9)
```

Stebalien:
@NobodyXu I agree if you're only target is Linux, but it's still useful to have a platform-independent way to do this (to, e.g., support MacOS which hasn't seen any filesystem API improvements in, basically ever).

src/lib.rs
```diff
@@ -399,6 +401,63 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+    /// `umask` applied by the underlying `open` syscall.
+    ///
+    /// ## Windows and others
+    ///
+    /// This setting is ignored.
+    ///
+    /// # Limitations
+    ///
+    /// Permissions for directories aren't currently set even though it would
+    /// be possible on Unix systems.
```
Stebalien, line 428: We need to implement this before we merge this patch, unfortunately. This builder is used to create temporary directories as well, and I don't really trust users to read the docs.

src/lib.rs
```diff
@@ -399,6 +401,63 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+    ///
+    /// # Platform Notes
+    /// ## Unix
+    ///
+    /// The actual permission bits set on the tempfile will be affected by the
+    /// `umask` applied by the underlying `open` syscall.
+    ///
+    /// ## Windows and others
+    ///
+    /// This setting is ignored.
```
Stebalien, line 423: This is annoying... but there's not much we can do here. Well, maybe check `.readonly()` on windows and fail if it's set?

src/lib.rs
Stebalien: See the discussion in #61.

Stebalien:
Thanks! And yes, I prefer the final version (more invasive approach).

src/lib.rs
Byron: I took a look at this and saw that `std::fs::create_dir()` is used to create the directory, I don't think the standard library has a way to handle this.
Did you have an idea on how to do this? Maybe a `rustix` `chmod` call?

src/lib.rs
Byron: That's a great idea! I have added this check in the most recent commit.

Byron:
Thanks for the quick review! If CI passes I have addressed the windows portion of it, but will definitely need some guidance to deal with the directory. If you have ideas and want to tinker, please feel free to push into this branch - sometimes that's easiest.

@NobodyXu Thanks for sharing! It's definitely something to consider for the time when performance needs to be optimised further. Personally I am most interested in doing what Git does at first, which seems to be pretty much what's happening here.

src/lib.rs
Stebalien: I think we'll need to use `rustix::fs::mkdir`.

Byron:
I have added mode-support for directories thanks to your hint with `rustix::fs::mkdir` and it appears to work. Further, trying to change permissions on unsupported platforms will result in an error that as well.
When CI is green it's ready for review, with the remark that I kept `dir.rs` nimble, with a structure sufficient for the task even though it's not quite the same as in `file`.

src/dir.rs
```diff
@@ -468,10 +468,76 @@ impl Drop for TempDir {
...
+        use std::io;
+        use std::os::unix::fs::PermissionsExt;
+        use std::path::PathBuf;
+
+        pub fn create(
+            path: PathBuf,
+            permissions: Option<&std::fs::Permissions>,
+        ) -> io::Result<TempDir> {
+            let mode: rustix::fs::Mode = permissions.map_or(
+                Mode::RWXU | Mode::RGRP | Mode::XGRP | Mode::ROTH | Mode::XOTH,
```
Stebalien, line 492: I believe the correct thing to pass here is 0777, and let the umask deal with it (we're being inconsistent with tempfile.... but changing that now isn't a great idea).

src/dir.rs
```diff
@@ -468,10 +468,76 @@ impl Drop for TempDir {
...
+                    |errno| {
+                        io::Error::new(
+                            errno.kind(),
+                            crate::error::PathError {
+                                path,
+                                err: io::Error::from_raw_os_error(errno.raw_os_error()),
+                            },
+                        )
+                    }
+                })
```
Stebalien, line 507: You can write this as:

```rust
                .map_err(Into::into)
                .with_err_path(|| &path)
```

Then revert the change to make the fields public.

src/dir.rs
```diff
@@ -468,10 +468,76 @@ impl Drop for TempDir {
...
-            path: path.into_boxed_path(),
-        })
+pub(crate) fn create(
+    path: PathBuf,
+    permissions: Option<&std::fs::Permissions>,
+) -> io::Result<TempDir> {
+    imp::create(path, permissions)
+}
+
+mod imp {
```
Stebalien, line 478: I'd prefer it if we created the entire directory structure.

Stebalien:
So... somehow I missed this but: https://doc.rust-lang.org/std/fs/struct.DirBuilder.html#method.mode. We don't need to use rustix here.

src/dir.rs
Byron: I can do that, even though it's a change that might have performance implications. For that reason alone I wouldn't do it without also introducing a new method for it, and do so in a separate PR.

src/dir.rs
Byron: With `std::fs::DirBuilder` we have the luxury of not specifying it unless it's overridden, I will do that.

Byron:
> So... somehow I missed this but: https://doc.rust-lang.org/std/fs/struct.DirBuilder.html#method.mode. We don't need to use rustix here.

Oh my, I missed that too! I clearly was biased when researching this topic and "didn't believe in mode being available for directories" 🤦‍♂️. Lesson learned, and the code was adjusted to something much better.

I made a note that recommends to not recursively create the directory by default but leave that to a separate PR along with a new flag in the builder maybe to control this, to keep the previous behaviour exactly the same. Right now, this still is the case.

If CI is green, this PR should be ready for re-review.

src/dir.rs
Stebalien: Wait, no. I mean: put the modules in different files/directories (`imp/...`). Sorry, I realize that that was confusing.

Byron: Ah, right, I could have guessed it. And it's done now. A note regarding naming: It feels like `any` should be `other`, but I decided not to do that as `other` means unsupported in the `file` folder, while here it is a catch-all.

src/error.rs
```diff
@@ -2,9 +2,9 @@ use std::path::PathBuf;
 use std::{error, fmt, io};
 
 #[derive(Debug)]
-struct PathError {
-    path: PathBuf,
-    err: io::Error,
+pub(crate) struct PathError {
+    pub(crate) path: PathBuf,
+    pub(crate) err: io::Error,
```
Stebalien, line 7: This change can be revered.

src/dir/imp/unix.rs
```diff
@@ -0,0 +1,21 @@
...
+    #[cfg(not(target_os = "wasi"))]
+    {
+        use std::os::unix::fs::{DirBuilderExt, PermissionsExt};
+        if let Some(p) = permissions {
+            dir_options.mode(p.mode());
+        }
+    }
+    dir_options
+        .create(&path)
+        .with_err_path(|| path.clone())
```
Stebalien, line 17:
```suggestion
        .with_err_path(|| &path)
```

(the path is cloned automatically if there's an error)

src/lib.rs
Stebalien: This comment is now out of date.

src/dir.rs
Stebalien: Hm. Yeah, I may go through and do a rename later but those are implementation details so.. meh.

Stebalien:
And with those changes, this should be good to go. Thanks!

Byron:
Thanks a lot for the QA and sorry for the sloppiness, at least 2 of 3 I should have caught.
But here we are, and I think that's the version that is ready to go.

Stebalien (approved):
❤️

Stebalien:
> Thanks a lot for the QA and sorry for the sloppiness, at least 2 of 3 I should have caught.

Don't be. That'll just make me feel worse about my PRs and that's why we have review 😆.

> PS: A new release with these changes would help, as that way everything can trickle downstream to StackedGit. Thanks again.

Will do.

PR #279, opened 2024-03-12
Update version in html_root_url

stoeckmann:
If tempfile documentation is referenced without being built locally, the URLs point to version 3.1.0.

Not a big deal, but it won't hurt to update the URL.


Stebalien:
Ah, nice catch. Can you change this to "https://docs.rs/tempfile/latest/tempfile/" so we don't have to do this again?

stoeckmann:
Updated

Stebalien:
Thanks!

stoeckmann:
I've looked around in other crates. They do not use latest, but most of them skip the trailing slash. So I haven't added it here.

Stebalien:
Anything that ends up pointing to the right place is fine with me, honestly.

PR #280, opened 2024-03-12
chore: fix doc url

stoeckmann:
It has to end with version (or latest).
Fixes ecb73745c552f9df53f8943ea80e9bebf51643a2.

Just checked. Otherwise the links are broken. Sorry for the second push into the previous branch.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,7 +160,7 @@
 #![doc(
     html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
     html_favicon_url = "https://www.rust-lang.org/favicon.ico",
-    html_root_url = "https://docs.rs/tempfile/latest/tempfile"
+    html_root_url = "https://docs.rs/tempfile/latest"
 )]
 #![cfg_attr(test, deny(warnings))]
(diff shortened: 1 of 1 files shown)


Stebalien:
Where is this broken? https://docs.rs/tempfile/latest/tempfile works just fine for me.

stoeckmann:
The automatically generated URLs would point to the wrong locations though, since they append tempfile to the given URL on their own. So it would be  https://docs.rs/tempfile/latest/tempfile/tempfile/...

Stebalien:
Ah, got it.

Stebalien:
Hm. Actually, do we even need this attribute? I think we can just get rid of the entire `doc` attribute.

stoeckmann:
It's not a requirement for tempfile itself, see https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#html_root_url

It helps if someone has tempfile as a dependency and runs `cargo doc --no-deps`. The generated HTML files link to the tempfile documents at the given root. If no such root is given, no links are created.

So, nothing really bad happens if it's gone.

Stebalien:
Ok, might as well fix it. Thanks!

PR #286, opened 2024-07-26
feat: add the ability to override the default temporary directory

### LINKED ISSUE #285: Set temp directory (override TMPDIR environment)
Now that `std::env::set_var` is unsafe and essentially unusable anymore, it would be great if tempfile provided a method to override the global temporary directory on UNIX.

Stebalien:
This adds a new `env` module with `override_temp_dir` and `temp_dir` functions.

- `temp_dir` will defer to `std::env::temp_dir` by default unless the temporary directory has been overridden.
- `override_temp_dir` allows the user to override the default temporary directory ONCE. Once this value has been set, it cannot be changed Care should be taken to ensure that the chosen directory is actually writable.

This API is designed for use by the application author when the application may run in an environment without a reliable global temporary directory (e.g., Android).

fixes #285

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -13,7 +13,7 @@ static DEFAULT_TEMPDIR: OnceLock<PathBuf> = OnceLock::new();
 /// Only the first call to this function will succeed. All further calls will fail with `Err(path)`
 /// where `path` is previously set default temporary directory override.
 ///
-/// NOTE: This function does not check if the specified directory exists and/or is writable.
+/// **NOTE:** This function does not check if the specified directory exists and/or is writable.
 pub fn override_temp_dir(path: &Path) -> Result<(), PathBuf> {
     let mut we_set = false;
     let val = DEFAULT_TEMPDIR.get_or_init(|| {
@@ -33,7 +33,7 @@ pub fn override_temp_dir(path: &Path) -> Result<(), PathBuf> {
 /// This function simply delegates to [`std::env::temp_dir`] unless the default temporary directory
 /// has been override by a call to [`override_temp_dir`].
 ///
-/// NOTE: This function does check if the returned directory exists and/or is writable.
+/// **NOTE:** This function does check if the returned directory exists and/or is writable.
 pub fn temp_dir() -> PathBuf {
     DEFAULT_TEMPDIR
         .get()
--- a/src/env.rs
+++ b/src/env.rs
@@ -6,23 +6,24 @@ use once_cell::sync::OnceCell as OnceLock;
 
 static DEFAULT_TEMPDIR: OnceLock<PathBuf> = OnceLock::new();
 
-/// Override the default temporary directory (defaults to [`std::env::temp_dir`]. This function
+/// Override the default temporary directory (defaults to [`std::env::temp_dir`]). This function
 /// changes the _global_ default temporary directory for the entire program and should not be called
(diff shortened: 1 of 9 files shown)


Stebalien:
(needs test)

src/env.rs
```diff
@@ -0,0 +1,40 @@
...
+static DEFAULT_TEMPDIR: OnceLock<PathBuf> = OnceLock::new();
+
+/// Override the default temporary directory (defaults to [`std::env::temp_dir`]. This function
+/// changes the _global_ default temporary directory for the entire program and should not be called
+/// except in exceptional cases where it's not configured correctly by the platform.
+///
+/// Only the first call to this function will succeed, returning `Ok(path)`. All further calls will
+/// fail with `Err(path)`. In both cases, `path` is the default temporary directory (on success, the
+/// new one; on failure, the existing one).
+pub fn override_temp_dir(path: &Path) -> Result<&'static Path, &'static Path> {
```
Stebalien, line 16: I considered treating this as a fallback (TMPDIR -> the override -> "/tmp"), but handling all the platform-specific cases got hairy.

I also considered testing if the various temporary directories were writable.... but honestly, I don't want to be in the position of "guessing" where temporary files should be written (although I'll revisit that later if this keeps coming up).

surban: I would suggest `Result<(), TempDirAlreadyOverridden>` as the return type.

In the success case there is not point in returning the value just passed in.
In the error case it is problematic that `Path` does not implement `Display`.

surban: Docs: Is there a requirement that the directory already exists (I think so) or will it be created (recursively) if non-existent?

surban (changes requested):

surban:
Generally I am not sure if it makes sense to limit the override function to be only called once. Is there a technical reason why you want to prevent it? 

For example, a system utility running during boot (initrd) might need to change its temporary directory after mounting some file systems.

src/env.rs
Stebalien: > In the success case there is not point in returning the value just passed in.

It returns a static reference, which could be convenient if you need to know where the temporary directory is. But... it's inconsistent with `temp_dir`, so it's probably not all that useful.

Stebalien:
I can relax the constraints later if someone presents a strong use-case, but tightening them isn't possible so I'm starting with the strongest constraints.

My reason is that `override_temp_dir` exists _only_ to handle the case where the OS doesn't set an appropriate temporary directory. It's a last-resort escape hatch. I don't want is multiple libraries calling `override_temp_dir`, then getting confused when their setting gets clobbered by some other library.

In general, applications needing to put specific files in specific directories should use `tempfile_in` and friends.

> For example, a system utility running during boot (initrd) might need to change its temporary directory after mounting some file systems.

Init systems re-exec themselves for this reason. Changing the tempdir at runtime likely isn't going to have the desired result (old temporary files will still live in the old location).

src/env.rs
Stebalien: Hm. Ok, I'm going to change it to return `Result<(), PathBuf>`, but I actually don't want the user to `unwrap()` this error, they need to handle it.

Stebalien:
Merging this for now. I'll release at the end of the week so we can continue our discussion until then (and possibly iterate on the design).

PR #296, opened 2024-08-10
Document default permissions for tempdirs

### LINKED ISSUE #284: Clarify Builder::permissions documentation on defaults for temporary directories
The [documentation](https://docs.rs/tempfile/3.10.1/tempfile/struct.Builder.html#method.permissions) notes:

> The permissions to create the tempfile or [tempdir](https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#method.tempdir) with. This allows to them differ from the default mode of `0o600` on Unix.

I read this as tempdirs defaulting to `0o600`, but in the implementation it looks like this default is only applied for tempfiles, not tempdirs?

(sidenote: it looks like there was a suggestion in https://github.com/Stebalien/tempfile/issues/114 to make the default 0700 for directories)

Imberflur:
fixes #284

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -334,7 +334,6 @@ impl<'a, 'b> Builder<'a, 'b> {
     }
 
     /// The permissions to create the tempfile or [tempdir](Self::tempdir) with.
-    /// This allows to them differ from the default mode of `0o600` on Unix.
     ///
     /// # Security
     ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -347,6 +346,8 @@ impl<'a, 'b> Builder<'a, 'b> {
     /// # Platform Notes
     /// ## Unix
     ///
+    /// Tempfiles default to `0o600` and temdirs default to `0o777`.
```
Stebalien, line 349: This should be combined with the next sentence for clarity and should explain how the umask is combined. The reader needs to walk away understanding that, by default, the permissions on a directory are `0o755` given the standard `0o022` umask, etc.

Also, it should mention `Builder::permissions`.

Stebalien: Oh, wait. This _is_ `Builder::permissions()`. Never mind that part.

But I still want the reader to come away with an understanding that their temporary directory is probably _not_ going to be world-writable.

Imberflur: I tried to incorporate this, but further notes on it would be welcome.

Stebalien (approved):
Thank you!

src/lib.rs
Stebalien: That's plenty good enough.

PR #299, opened 2024-09-28
plumbing for `with_suffix` for issue #298

### LINKED ISSUE #298: feature: `with_suffix` to correspond to `with_prefix`
Stumbled upon while trying to wrangle some integration testing for a CLI tool, which expects a specific filetype. `with_prefix` adds a short string to the start of a file's name. We can alter this very cleanly by simply doing some plumbing for adding a short string to the end of the file's name, instead. `Builder` already has a `suffix` method which we can leverage for this.

Borgerr:
closes #298 

Adds a `with_suffix` to correspond to `with_prefix`. Allows for easier filetype specification and naming.

### DIFF
--- a/src/dir/mod.rs
+++ b/src/dir/mod.rs
@@ -267,6 +267,58 @@ impl TempDir {
         Builder::new().prefix(&prefix).tempdir()
     }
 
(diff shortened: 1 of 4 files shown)


Stebalien (approved):
Thanks!

PR #304, opened 2024-11-04
allow older windows-sys 0.52

king-11:
dependabot bumped the version to 0.59 but older version 0.52 can also be supported allowing downstream crates to pull dependency as per their version choice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ once_cell = { version = "1.19.0", default-features = false, features = ["std"] }
 rustix = { version = "0.38.37", features = ["fs"] }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.59"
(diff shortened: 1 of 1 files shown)


Stebalien:
Why?

king-11:
I don't think tempfile makes use of any API change which might have been introduced with `0.59` so keeping the range would prevent dependency hell for other downstream dependencies to only pull one version and I have mostly found that to be version `0.52`.

Stebalien:
Hm. There's no guarantee that'll work as you expect (https://github.com/rust-lang/cargo/issues/10599) but it's unlikely to hurt given that this dependency isn't exported.

Stebalien (approved):

king-11:
Yep the resolver can't really make out of this NP hard problem but many a times it works and gets the right one.

Stebalien:
3.14 published

PR #305, opened 2024-11-05
fix: enable wasip2 feature for wasm32-wasip2 target

cdmurph32:
- wasip2 will require +nightly until https://github.com/rust-lang/rust/issues/130323 is resolved and/or std::os::wasip2 is available in stable.
- Support was added to rustix for version 0.38.39 https://github.com/bytecodealliance/rustix/pull/1205

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,18 @@ jobs:
           targets: wasm32-wasip1
       - name: Build
         run: cargo build --target wasm32-wasip1 --features nightly
+  wasip2:
+    name: WASI P2 Test Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          targets: wasm32-wasip2
+      - name: Build
+        run: cargo build --target wasm32-wasip2 --features nightly
   wasm:
     name: WASM Test Build
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,4 +37,3 @@ doc-comment = "0.3"
 
 [features]
 nightly = []
-unstable = []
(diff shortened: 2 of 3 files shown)


Stebalien:
From my reading of that issue, this isn't the correct fix. See https://github.com/servo/rust-url/pull/983.

Am I missing something?

cdmurph32:
In that particular case, @brooksmtownsend was able to get around the use of std::os::wasi::prelude::OsStrExt. 

This PR allows for the use of rustix for the wasm32-wasip2 target.  There probably isn't a way to have this crate work for that target without rustix for the numerous uses of `std::os::wasi` without reinventing the wheel.

brooksmtownsend:
Yeah just to add a bit of context, my original PR to "url" was to enable compilation when wasm32-wasip2 was only on the nightly channel, then my follow up PR was to remove the use of those unstable APIs by using essentially equivalent APIs.

Put another way, this crate currently does not compile to wasm32-wasip2. This PR would enable compilation on the nightly channel, and then enabling compilation on stable would either require a Rust change in the feature gating or to replace the use of those unstable APIs

Stebalien:
I see... Let's put this behind a new `unstable` feature flag in `tempfile`. It sounds like this is liable to break at any time and I don't want users to be surprised.

Stebalien:
I forgot all about the existing "nightly" feature. I've pushed a change that merges "unstable" into "nightly", does this work for you?

cdmurph32:
Looks good. Builds on my machine. 😄

Stebalien (approved):

Stebalien:
Hm. It looks like cargo-wasi was just broken by https://github.com/rust-lang/compiler-team/issues/607.

cdmurph32:
`cargo wasi` is deprecated. Just use `cargo build --target wasm32-wasip2` for this change.

Stebalien (approved):

Stebalien:
3.14 published

PR #309, opened 2024-11-13
docs: add owasp link on insecure temporary files

n0toose:


n0toose:
I nearly created an insecure implementation relying on temporary files. Despite the (very) useful reminder in the documentation itself, I think it would also make sense to provide a more concrete explanation on the subject in general, even if it doesn't directly apply to Rust or the crate itself.

Stebalien (approved):

Issue #303, opened 2024-10-27
Mention `override_temp_dir` and custom permissions on the main crate documentation page


erickguan:
I propose updating the front page of the crate documentation to include brief mentions of key features that are currently documented but might not be immediately visible to users. Specifically:

* Programmatically setting the temporary directory using `tempfile::env::override_temp_dir`
* Setting custom permissions for temporary files and directories using the `Builder` pattern

Motivation:

While these features are already detailed elsewhere in the documentation, mentioning them upfront would:

* Improve visibility: New users can quickly become aware of these capabilities.
* Promote best practices: Encourages developers to consider security and control from the start.
* Enhance usability: Makes the crate’s features more accessible and easier to discover.

Draft of a new section or as new examples:

**Set custom permissions with the `Builder` pattern**

Create temporary files or directories with custom permissions to enhance security.

```rust
use tempfile::Builder;
use std::fs::Permissions;
use std::os::unix::fs::PermissionsExt;

// Create a temporary directory with 0o700 permissions
let temp_dir = Builder::new()
    .permissions(Permissions::from_mode(0o700))
    .tempdir()?;

// The directory is now only accessible by the owner
```

**Programmatically set the temporary directory**
You can override the default temporary directory within your application without relying solely on environment variables.

```rust
use tempfile::env;
use std::path::PathBuf;

// Override the temporary directory. read more in `env`.
env::override_temp_dir(PathBuf::from("/custom/temp/dir"));

// Now, any temporary files or directories will be created in "/custom/temp/dir"
let temp_file = tempfile::NamedTempFile::new()?;

// explicitly check if we deleted the file successfully. read more in `NamedTempFile`.
temp_file.close()?;
```

Stebalien:
These are two different things:

1. `mkdtemp` just creates a temporary directory with a mask of 0o700, you can currently do that with `Builder::new().permissions(Permissions::from_mode(0o700)).tempdir()`.
2. In terms of picking a secure directory, `tempfile` does what you tell it to do. Ideally you'll have configured `TMPDIR` correctly but, if not, it's possible for an application to override it with `tempfile::env::override_temp_dir`. E.g., an application can create a temporary directory then set the application-wide temporary directory with 0o700 permissions to be that new temporary directory.

erickguan:
Thanks so much for the quick reply!

1. You make an excellent point about the `Builder` and how it already allows setting a secure mask with Permissions::from_mode(0o700). I hadn't considered that. I agree with you.

[...]

PR #311, opened 2024-11-17
docs: improve security documentation

n0toose:
Minor wording fixes. Increases visibility of `Builder::permissions`,
the `NamedTempFile` Security documentation and
`env::override_temp_dir`.

Partially addresses #303.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,10 +26,13 @@
 //! In the presence of pathological temporary file cleaner, relying on file paths is unsafe because
 //! a temporary file cleaner could delete the temporary file which an attacker could then replace.
 //!
-//! `tempfile` doesn't rely on file paths so this isn't an issue. However, `NamedTempFile` does
-//! rely on file paths for _some_ operations. See the security documentation on
-//! the `NamedTempFile` type for more information.
-//! 
+//! `tempfile` doesn't rely on file paths, so this should not be an issue. However,
+//! `NamedTempFile` does rely on file paths for _some_ operations. On Unix-like
+//! operating systems and depending on your use case, it may be possible to
+//! mitigate this issue by overriding the crate's default options. For more information,
+//! consult the Security documentation of the [`NamedTempFile`] type,
+//! [`Builder::permissions`] and [`env::override_temp_dir`].
+//!
 //! The OWASP Foundation provides a resource on vulnerabilities concerning insecure
 //! temporary files: https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File
 //!
@@ -172,7 +175,7 @@ pub use crate::file::{
 };
 pub use crate::spooled::{spooled_tempfile, SpooledData, SpooledTempFile};
 
-/// Create a new temporary file or directory with custom parameters.
+/// Create a new temporary file or directory with custom parameters and permissions.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Builder<'a, 'b> {
     random_len: usize,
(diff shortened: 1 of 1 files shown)


n0toose:
I tried to avoid rewriting the text, give specific instructions in the first page of the documentation (so that they won't be implemented without nuance) and instead opted for pointing people to the right parts of the documentation - where we have enough space to explain everything precisely - and an external resource that explains the problem to the user.

Every use case is different, and figuring out the security aspects should be the user's responsibility - accidentally misleading the user is not very worth it.

n0toose:
This also fixes an unfortunate `cargo fmt` problem from my previous PR.

src/lib.rs
```diff
@@ -172,7 +175,7 @@ pub use crate::file::{
 };
 pub use crate::spooled::{spooled_tempfile, SpooledData, SpooledTempFile};
 
-/// Create a new temporary file or directory with custom parameters.
+/// Create a new temporary file or directory with custom parameters and permissions.
```
Stebalien, line 178: IMO, there's no need to elaborate here. Parameters imply permissions, etc. (could also just call it "options" if you find that less confusing).

src/lib.rs
```diff
@@ -26,10 +26,13 @@
...
-//! `tempfile` doesn't rely on file paths so this isn't an issue. However, `NamedTempFile` does
-//! rely on file paths for _some_ operations. See the security documentation on
-//! the `NamedTempFile` type for more information.
-//! 
+//! `tempfile` doesn't rely on file paths, so this should not be an issue. However,
+//! `NamedTempFile` does rely on file paths for _some_ operations. On Unix-like
+//! operating systems and depending on your use case, it may be possible to
+//! mitigate this issue by overriding the crate's default options. For more information,
+//! consult the Security documentation of the [`NamedTempFile`] type,
+//! [`Builder::permissions`] and [`env::override_temp_dir`].
```
Stebalien, line 34: Note: `Builder::permissions` is relevant for `Builder::tempdir`, but not for `Builder::tempfile` (`NamedTempFile`) because:

1. Changing the permissions of the file won't stop temporary file cleaners (the main issue).
2. The permissions default to 0600.

However, setting the permissions _is_ an important security consideration for temporary _directories_, so that should probably be called out.

Stebalien (commented):
Thanks!

src/lib.rs
n0toose: > The permissions default to 0600.

I presume that this is not the case on Windows (with an equivalent permission)? I sent a revised version.

n0toose: Either way, it would be interesting to document that so that people know how to work around that if necessary in the meantime (e.g. using `/tmp` on Linux, but using a custom directory with specific permissions on Windows) and to allow room for a contribution that fixes the need for such a workaround to be made.

n0toose: Thank you for clarifying. I think I did not really see the forest for the trees. I'll revise my change again for increased clarity.

> a temporary file cleaner could delete the temporary file which an attacker could then replace.

n0toose: I changed the wording re: "predictable file names" without explicitly explaining the problem. I chose to link to the affected / relevant functions instead.

Stebalien:
While reviewing this I came to the conclusion that the security documentation needed to be fleshed out into multiple sections, so I wrote https://github.com/Stebalien/tempfile/pull/317 which _should_ address the concerns in this PR. Unfortunately, that took a while because I realized I could just "fix" the DoS vector (https://github.com/Stebalien/tempfile/pull/314) in a way that was unlikely to annoy too many people.

Could you give #317 a review?

n0toose:
> Could you give https://github.com/Stebalien/tempfile/pull/317 a review?

Sure thing! Thank you for the explanation btw. As a contributor, I really appreciate it when maintainers explain their rationale behind decisions like this, even when they're not obliged to do so.

Stebalien:
:heart: thanks!

PR #317, opened 2025-01-03
docs: improve & expand security documentation

Stebalien:
Alternative to #311 (this patch includes a few miscellaneous documentation fixes from that PR as well).

The previous documentation didn't sufficiently cover permissions issues and mixed all the security concerns into a single section. This patch separates things out into separate sections and hopefully makes all this easier to understand.

This patch also documents the DoS mitigation introduced in #314 (I stared writing this documentation first then decided that I'd rather just fix the DoS vector...).

I've also removed the link to the OWASP documentation to avoid confusing users (their documentation is mostly concerned with low-level C and platform-specific temporary file creation functions).

### DIFF
--- a/src/dir/mod.rs
+++ b/src/dir/mod.rs
@@ -22,14 +22,19 @@ use crate::env;
 
 /// Create a new temporary directory.
 ///
-/// The `tempdir` function creates a directory in the file system
-/// and returns a [`TempDir`].
-/// The directory will be automatically deleted when the `TempDir`s
+/// The `tempdir` function creates a directory in the file system and returns a
+/// [`TempDir`]. The directory will be automatically deleted when the `TempDir`'s
 /// destructor is run.
 ///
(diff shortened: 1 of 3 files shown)


n0toose (approved):
This is straight to the point and does not cram everything into one place. I definitely prefer this over https://github.com/Stebalien/tempfile/pull/311. It provides enough context for the implementer to decide what works best for them.

n0toose:
P.S. I used your documentation as a point of reference today, as I was reviewing whether my usage of your crate was sane. I think I have to mention that it was *immensely* helpful from my perspective. I found the explanations in the sections "Denial of Service" and "Temporary File Cleaners" to be very, very good. I really liked the step-by-step Mitigations and the short explanations as to whether "this problem is likely to affect the developer in any capacity whatsoever".

Stebalien:
I'm glad it was helpful!

PR #319, opened 2025-01-21
build: allow `windows-sys` 0.59

ErichDonGubler:


Cargo.toml
```diff
@@ -29,7 +29,7 @@ getrandom = { version = "0.2.15", default-features = false, optional = true }
 rustix = { version = "0.38.39", features = ["fs"] }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = ">=0.52,<=0.59"
+version = ">=0.52,<=0.60"
```
Stebalien, line 32: The current version selection should work with 0.59 (the current latest version), although it'll have issues with 0.59.1 etc. This should have been `>=0.52,<0.60.0` (strictly less than).

Stebalien (changes requested):

Cargo.toml
ErichDonGubler: olol, I've been writing so many of these bump PRs that I didn't even notice the equals already there. Sigh! Good catch. 😅

Stebalien (approved):
Thanks!

ErichDonGubler:
@Stebalien: Wow, that was fast! Thanks! Excited to see this land on Crates.io. Were you thinking that would be soon, or should we expect a longer wait time for that?

Stebalien:
Do you need a release? 0.59.0 is currently allowed, just not 0.59.1 etc. (which don't exist anyways).

ErichDonGubler:
@Stebalien: Putting on my maintainer hat, I'd recommend a release just to flush the issue out to a published version, but I don't have a hard blocker right now for the work that motivated me to make this PR. 🫡 Do as you see fit!

Stebalien:
I just cut a release and don't like cutting tiny releases unless necessary (time, noise, etc.).

PR #320, opened 2025-01-25
build: upgrade getrandom to v0.3.0

paolobarbolini:
This upgrades to `getrandom` v0.3, which was released today.
Changelog: https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#030---2025-01-25

It may make sense to wait a bit before releasing this upgrade as the rest of the ecosystem (including `rand`) hasn't upgraded yet.


Stebalien (approved):
Thanks! As you suggest, I'll hold off on releasing for a bit. But I'll get this merged so I don't forget about it later.

PR #323, opened 2025-02-08
fix: always use absolute paths

Stebalien:
We used absolute paths in the builder when making directories and in `create_named` but this didn't cover `Builder::make_in` and technically introduced a small race in some unnamed tempfile cases (i.e., when we can't atomically create unnamed temporary files and need to create then immediately delete them).

Instead, we now resolve the filename into an absolute path inside the main create helper.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -1043,16 +1043,11 @@ impl<F: AsRawHandle> AsRawHandle for NamedTempFile<F> {
 }
 
 pub(crate) fn create_named(
-    mut path: PathBuf,
+    path: PathBuf,
     open_options: &mut OpenOptions,
     permissions: Option<&std::fs::Permissions>,
     keep: bool,
 ) -> io::Result<NamedTempFile> {
-    // Make the path absolute. Otherwise, changing directories could cause us to
-    // delete the wrong file.
-    if !path.is_absolute() {
-        path = std::env::current_dir()?.join(path)
-    }
     imp::create_named(&path, open_options, permissions)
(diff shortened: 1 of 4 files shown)


Stebalien:
Hm. The issue is that tempfile needs to know where exactly the file lives so it can correctly delete it. In your case, it would have attempted to delete `filename` in the current directory, ignoring `dirfd`.

I'm curious, what's your use-case? I've considered supporting directory-relative temporary files (https://github.com/Stebalien/tempfile/issues/40) but I've never had a strong use-case.

nagisa:
Aha, nicely spotted. Yeah, deletion would have probably been buggy too. The only reason I hadn't hit it was because in the happy path all of these temporary files would get retained (`into_temp_path()` + `renameat2`) and `renameat2` failure would've been the only time deletion logic would've triggered.

To elaborate more on the overarching use-case: my code is implementing a filesystem-level cache. One of the concerns are that the lookups for a given filename should only succeed once the value has been fully written (hence the `rename(at2)`.) And the reason for `dirfd` use was primarily due to not wanting to deal with all possible corner cases of the meaning of base directory path changing. This in my mind seemed like a significant enough concern where the binary might run for months at a time and `rm -rf everything` being plausible enough (it isn't invalid to remove the files within the directory but I didn't want to deal with the *directory* itself getting nuked.)

To be honest I don't mind at all implementing all of that funky logic myself, `tempfile` came in as a dependency, as I mentioned, chiefly to avoid implementing the somewhat subtle random name generation code and instead focus on the subtlety in the other parts of the logic there ^^

Issue #258, opened 2023-10-11
[META] Pain Point Tracker


Stebalien:
This is where I'm going to keep a list of known pain points with tempfile v3 that can't be fixed without breaking changes. I don't know if we'll ever _have_ a v4, but if we do, we can revisit everything on this list:

- [`TempDir` implements `AsRef<Path>`](https://github.com/Stebalien/tempfile/issues/115) making it easy to accidentally drop it.
- [A file name prefix can contain a slash](https://github.com/Stebalien/tempfile/issues/59). May be fixable in a minor version with a warning?
- [Temporary directories are created with the default mode unlike temporary files which are created with 0o600](https://github.com/Stebalien/tempfile/issues/284). Ideally temporary directories would use restrictive permissions by default.
- The builder has two lifetime parameters making it awkward to add additional fields with new lifetime parameters. Ideally, it would just have a single lifetime and we'd rely on subtyping.
- Take shared advisory locks on all named temporary files to prevent temporary file cleaners from deleting them. This could break some things so we have to be careful, but taking a shared lock is probably reasonable? Possibly only do this if the parent directory is `+t` (sticky)?

marcospb19:
I'd expect the #59 fix to require a major version update, it's not obvious to users whether that's a bug or intended behavior, so that should be assumed as breaking IMHO.

Stebalien:
Closing in favor of https://github.com/Stebalien/tempfile/pull/327 where I'll be tracking all 4.0 changes.

Issue #59, opened 2018-05-25
Update `Builder` (or `create_helper`) to check that the supplied `prefix` and `suffix` don't contain `/`


cipriancraciun:
As the title says, currently the code doesn't make any validations if the `prefix` or `suffix` are valid path component values (i.e. they don't contain a slash or `\0`).

(If needed I can provide the patch.)

Stebalien:
So, filtering `\0` shouldn't be necessary as the OS will do that. Really, we can't reasonably build a blacklist for every OS so we might as well delegate.

In retrospect, we probably should have checked for path separators. However, the documentation currently _explicitly_ allows them.

cipriancraciun:
@Stebalien As explained in one of the previous issues, my main usage for this library is in a runtime for a (shell) scripting language, and thus many of the parameters sent to `tempfile` could be under the direct control of the "user" (which sometimes can be an "attacker")...

Now I understand that all developers should do their best in sanitizing the user inputs, before passing these values to the underlying libraries.  However this imposes extra burden on those developers (which is the first reason they use a library and not write their own).  Moreover in order to "perfectly" sanitize the inputs, one must have in-depth knowledge of how said library would use these inputs, which again adds development overhead.

Imagine one uses Rust and the `tempfile` library to create an application that has the `set UUD` bit set (i.e. runs as `root`), and fails to sanitize some prefix or suffix (which are exposed in one way or another as arguments) not to contain any slashes.  Now an "attacker" could just try to miss-use these arguments to create files outside the "temporary" directory (and target for example the `/etc/cron.d` folders).  (There are countless CVE's that follow this pattern in which an application that happens to run as root fails to sanitize inputs, and thus exposes the system to all kinds of issues.)

Therefore, especially since `tempfile` is so popular, I think the library should do its best to enforce some basic "sanity checkups".

Stebalien:
I agree but `Builder::new().prefix("foo/bar").tempfile()?` is valid and potentially useful so I'm worried someone is already using it. However, I really do agree that we _shouldn't_ allow that and users should just use `tempfile_in`.

@KodrAus thoughts? I'm really tempted to fix this as a security oversight.

KodrAus:
@Stebalien Personally, I think changing the behaviour here to prevent the prefix and suffix from altering the path is reasonable. I wouldn't necessarily expect `.prefix("foo/bar")` to do that or try use it that way myself in any case.

marcospb19:
[...]

PR #327, opened 2025-02-15
Tempfile V4

Stebalien:
This is a stack of changes for tempfile v4. My plan is to wait for the next Debian stable release then set the MSRV to that, although I may merge and cut an RC before then. This version tries to be minimally breaking in "normal" code but it will still break some code (ideally code that was already broken/buggy).

Other than the MSRV bump (letting us drop `once_cell`), changes include:

- Implement `AsRef<Path>` on `&NamedTempFile` (and friends) instead of directly on `NamedTempFile`. That way, functions accepting `AsRef<Path>` won't be able to take (then drop) `NamedTempFile` by value. See https://github.com/Stebalien/tempfile/issues/115.
- Temporary directories are created with 0700 permissions by default instead of 0755 to match temporary files.
- Temporary filenames no longer start with a `.` by default. Temporary files shouldn't generally be hidden. Additionally, it's now possible to configure the default temporary-filename prefix via the `env` package.
- The `Builder` can now has a single lifetime instead of separate ones for `prefix` and `suffix`. It can also be constructed with a const `Builder::new` if desired.
- Temporary file names are validated to be valid file-*names*, not including path separators, etc. It's still not safe to allow an attacker to control the prefix/suffix, but this gives us a little extra safety.
- The closure passed to `Builder::make` used to take a `&Path` but now takes an `&MakeParams` struct (that dereferences to a `&Path`). This lets us pass additional builder parameters like the desired permissions.
- The deprecated `TempDir::into_path` is removed in favor of `TempDir::keep`.
- `SpooledTempDir::new` is now const.
- `tempfile::env::temp_dir()` now returns a result and won't panic on platforms where `std::env::temp_dir` would panic.
- Unnamed temporary files may now be created in inaccessible directories (directories where we're missing search permissions on one of the parent directories).

TBD:

- [ ] Prevent temporary file cleanup by taking a shared advisory locks on all named temporary files to prevent temporary file cleaners from deleting them. This could break some things so we have to be careful, but taking a shared lock is probably reasonable? Possibly only do this if the parent directory is `+t` (sticky)? We can also make it configurable in the builder, the question will be: what's the default behavior?

Fixes:

- #59 
- Everything in #258
- #40


Imberflur:
One minor annoyance I noticed with `TempDir::close` is that you no longer have access to the path for error reporting, so it has to be copied beforehand.

NobodyXu:
debian trixie has released with rust 1.85, though IIRC one of the other popular distro (centos or fedora, I can't remember) still on 1.84

Stebalien:
> though IIRC one of the other popular distro (centos or fedora, I can't remember) still on 1.84

Ah, that's good to know. I'll see if I can downgrade to that.

Also note, I'm still planing on maintaining v3 for the foreseeable future.

Stebalien:
It's Ubuntu plucky. Which is unfortunately rules out 2024 edition unless I wait until October (although it might take that long anyways if I try to tackle the remaining TODOs).

Stebalien:
This now bumps the MSRV to 1.85 and the edition to 2024.

Stebalien:
This is now blocked on:

1. A libc release for `openat` on RedoxOS
2. https://github.com/bytecodealliance/rustix/pull/1594 (blocking updating libc in rustix).
3. A PR in rustix to implement `openat` and `unlinkat` for RedoxOS.
4. A rustix release.

PR #345, opened 2025-04-02
run `wasm32-wasip1` and `wasm32-wasip2` tests on CI

### LINKED ISSUE #326: Run WASI tests in CI
We check that we can build against wasip1 and wasip2, but don't run our tests on these platforms.

folkertdev:
fixes #326 

Actually running the tests is straightforward (well, we've figured this out before in zlib-rs), but most tests fail because the `std::env::temp_dir()` function panics for the wasi platforms (see https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/wasi/os.rs#L267).

So, I'm using `override_temp_dir` in combination with the `--dir "/tmp"` argument to wasmtime to make the tests pass. That's a bit annoying, but the best I could come up with. There are also some tests that are ignored now for the wasi targets because e.g. they spawn threads and that is just not supported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,14 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           targets: wasm32-wasip1
-      - name: Build
-        run: cargo build --target wasm32-wasip1 --features nightly
+      - name: Install wasmtime
+        run: curl https://wasmtime.dev/install.sh -sSf | bash
+      - name: Test
+        env:
+          RUST_BACKTRACE: 1
+          RUSTFLAGS: ""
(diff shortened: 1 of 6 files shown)


Stebalien:
Thank you! I've been sitting on this trying to figure out a good way to make this "sane by default" (or at least not panic) but I think anything I do there will require a v4 (at the very least, to break the `env::` APIs).

Unfortunately, there's no "before test" hook in rust so I can't even avoid the testing noise.

I've decided to merge this anyways because it'll help me make changes without having to second-guess how it might impact platforms like wasm.

Stebalien (approved):

Stebalien (approved):

PR #348, opened 2025-05-07
Build docs.rs documentation with rustdoc jump to def feature enabled

GuillaumeGomez:
I was reading source code on docs.rs and realized that the feature was not enabled, so here it is. :)

Enabling this feature makes your source code pages generate links on items so you can jump to where they're defined.

You can this feature in action in [`syn`](https://docs.rs/syn/latest/src/syn/lib.rs.html#986-1010) for example.


Stebalien:
Thanks!

PR #355, opened 2025-05-11
feat: make it possible to disable cleanup in-place after creation

### LINKED ISSUE #347: Conditional keep for TempDir after initialization
Hi!

I was working with some utilities on top of tempfile and I ran into an interesting issue: in some cases (e.g. tests) it's often useful to mark something as conditionally persistent (e.g. if a debugging environment variable is set).

I'm definitely aware of the ability to persist `TempDir`s via `into_path` (as discussed in #315), but that changes the type from `TempDir` to `PathBuf`. That's useful in some cases but gets in the way of others (it's annoying to have to manage your own enum just for this).

Would it be possible to expose a `keep` method, just like there is for `NamedTempFile`? Having such a method would aid this use case tremendously.

Thanks!

Stebalien:
This makes it possible to disable automatic cleanup of temporary files & directories _after_ they've been created.

1. Rename `Builder::keep` into `Builder::disable_cleanup` for consistency. `::keep` on `NamedTempDir`, `TempDir`, `TempPath`, etc. provide a way to _convert_ (one-way) a temporary directory/file into a persistent temporary directory/file but aren't a way to simply disable automatic cleanup in-place.
2. Add `disable_cleanup` methods to all named temporary file types, making it possible to replicate the behavior of `Builder::disable_cleanup` _after_ creating a temporary file. Given that this operation is in-place, it can be undone.

fixes #347

### DIFF
--- a/src/dir/imp/any.rs
+++ b/src/dir/imp/any.rs
@@ -10,7 +10,7 @@ fn not_supported<T>(msg: &str) -> io::Result<T> {
 pub fn create(
     path: PathBuf,
     permissions: Option<&std::fs::Permissions>,
-    keep: bool,
+    disable_cleanup: bool,
 ) -> io::Result<TempDir> {
     if permissions.map_or(false, |p| p.readonly()) {
         return not_supported("changing permissions is not supported on this platform");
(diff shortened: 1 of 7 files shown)


sunshowers:
This looks fantastic. Thank you!

PR #357, opened 2025-05-11
ci: really check docs for warnings

Stebalien:


Stebalien:
I moved too fast on the last one.

PR #359, opened 2025-05-12
docs: Remove mention of deprecated function

alexpovel:
`keep` is marked as deprecated in favor of `disable_cleanup` - so `disable_cleanup` should not recommend using `keep`.

Not totally sure this is how it's supposed to be used (as these docs are on `Builder`), but in any case the docs seem cyclic.

Let me know what you think!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -466,8 +466,7 @@ impl<'a, 'b> Builder<'a, 'b> {
     }
 
     /// Disable cleanup of the file/folder to even when the [`NamedTempFile`]/[`TempDir`] goes out
-    /// of scope. Prefer [`NamedTempFile::keep`] and `[`TempDir::keep`] where possible,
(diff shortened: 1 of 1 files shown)


alexpovel:
Noticed this as part of https://github.com/alexpovel/srgn/pull/206 .

Stebalien:
`Builder::keep` is deprecated and renamed to , `NamedTempFile::keep` is not and is preferred.

`Builder::keep` was deprecated (renamed to `Builder::disable_cleanup`). because it _sounded_ like it behaved like `NameTempFile::keep` but it was subtlly different.

alexpovel:
Thanks, I see! In that case, feel free to close this PR.

Stebalien:
(I should have proof-read that comment before posting it. That was.... gnarly, sorry about that.)

PR #364, opened 2025-07-13
Avoid to call env::temp_dir on wasi targets

Brooooooklyn:


Stebalien:
Also, do you have any links to any discussions I can read?

Brooooooklyn:
My bad, didn't realize this was not a standard implementation

Stebalien:
I honestly don't know. I've filed https://github.com/Stebalien/tempfile/issues/365 to consider a fix for the next major version, at least.

PR #377, opened 2025-09-16
doc: Update COPYRIGHT link

reddevilmidzy:
The previous link is no longer valid. After checking the [Wayback Machine](https://web.archive.org/web/20150124154549/https://www.rust-lang.org/COPYRIGHT), I found that its contents are similar to https://github.com/rust-lang/rust/blob/master/COPYRIGHT

### DIFF
--- a/src/dir/mod.rs
+++ b/src/dir/mod.rs
@@ -1,6 +1,6 @@
 // Copyright 2015 The Rust Project Developers. See the COPYRIGHT
 // file at the top-level directory of this distribution and at
-// http://rust-lang.org/COPYRIGHT.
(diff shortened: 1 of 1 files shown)


Stebalien:
From what I can tell, that entire (second) sentence can be cut. It's left over from when this file was a part of the rust project itself. At the time, they needed a separate file to explain the different copyright status/licenses of all the files in the repo.

In this case:

1. This _one_ file is copyright the rust project developers (already stated at the top).
2. This file (and the project) are all licensed under the same MIT OR Apache-2.0 license.

Stebalien:
Thanks!

Issue #385, opened 2025-12-23
[CI] FreeBSD/OpenBSD (NetBSD?)


Stebalien:
We don't currently have have any BSD CI and we really should.

PR #387, opened 2026-01-05
ci: adds FreeBsd ci using vm runner

ikilledkanishk:
Adds FreeBSD CI using a VM-based runner.

Addresses #385 by providing build and test coverage on FreeBSD,
unblocking platform-specific changes.


ikilledkanishk:
ci was failing because the freebsd vm doesn't have ssh installed by default, i have updated the workflow so it does now.

PR #401, opened 2026-04-27
Replace `windows-sys` with inline bindings

Jake-Shadle:
I'm going through dependencies on my projects attempting to remove as many usages of windows-sys as I can as the number of versions being pulled in is getting out of hand. This crate's needs are quite modest so I just inlined the bindings it needed to remove the dependency entirely.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,13 +29,6 @@ getrandom = { version = ">=0.3.0, <0.5", default-features = false, optional = tr
 [target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 rustix = { version = "1.1.4", features = ["fs"] }
 
-[target.'cfg(windows)'.dependencies.windows-sys]
-version = ">=0.52, <0.62"
-features = [
-    "Win32_Storage_FileSystem",
-    "Win32_Foundation",
-]
-
 [dev-dependencies]
 doc-comment = "0.3"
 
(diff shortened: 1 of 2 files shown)


Stebalien:
> I'm going through dependencies on my projects attempting to remove as many usages of windows-sys as I can as the number of versions being pulled in is getting out of hand.

In this case, I recommend you work with those projects to increase their range of supported `windows-sys` crates. `tempfile` allows any version from 0.52.x.y through 0.61.x.y so I doubt that's contributing to your crate pulling in multiple `windows-sys` versions. I can probably add support for even older versions, if that would help.

---

I'm inclined to leave the dependency because I don't use windows and testing/developing the Windows support in this crate is already tricky enough.

Usually, I'd consider removing a dependency if:

1. It's a niche dependency. Niche dependencies tend to receive less attention and can be juicy targets for attackers, especially when they're dependencies of widely used packages. However, `windows-sys` is one of the most popular crates on crates.io.
2. I don't know/trust the maintainer. I'm not a fan of Microsoft's software, but they'd lose a lot of cred if `windows-sys` got backdoored.
3. It added too much runtime bloat. However, in this case, I _assume_ the crate adds virtually no bloat given that it's just syscall stubs.

The one reason _to_ consider removing `windows-sys` is the cost of vendoring it. I understand that there's a lot of code there (even if it's inactive) and auditing it is likely a bit time consuming. On the other hand... it doesn't get updated that often and I'm skeptical any organization shipping rust code on windows will be able to completely remove it from their dependency tree.

Jake-Shadle:
Fair enough, I'm not under any illusions that I'll be able to remove it completely, I just got annoyed enough this week that I wanted to see how many crates I could remove it from. Thanks for taking the time to reply!

PR #403, opened 2026-05-06
make test_temp_path_resolve_existing_cwd() work with symlink in TMPDIR

martinvonz:
Here's a simple way to reproduce a test failure before this commit:

```
ln -s /tmp /tmp/here
TMPDIR=/tmp/here cargo test
```

That fails something like this:

```
test test_temp_path_resolve_existing_cwd ... FAILED

failures:

---- test_temp_path_resolve_existing_cwd stdout ----

thread 'test_temp_path_resolve_existing_cwd' (137617) panicked at tests/namedtempfile.rs:376:5:
assertion `left == right` failed
  left: "/tmp/.tmpDu1A7l/foo"
 right: "/tmp/here/.tmpDu1A7l/foo"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This commit fixes it by canonicalizing the CWD path in the test.

### DIFF
--- a/tests/namedtempfile.rs
+++ b/tests/namedtempfile.rs
@@ -363,13 +363,10 @@ fn test_temp_path_resolve_existing_cwd() {
     let tmpdir = tempdir().unwrap();
     std::env::set_current_dir(&tmpdir).expect("failed to change to directory");
 
-    let cwd = if cfg!(target_os = "macos") {
-        // MacOS has absolute paths and ABSOLUTE paths. `cd /var/tmp/...` actually changes to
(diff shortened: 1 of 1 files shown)


Stebalien:
Windows was failing because paths were getting converted to UNC syntax. So, instead of using the normalized paths in the rest of the tests, I just used the return value of `std::env::current_dir` and verified that we successfully changed to the correct directory.

I'm not super happy with it but... it works and it's just a test.

Stebalien:
But thanks for the PR and the easy reproducer!

PR #427, opened 2026-06-30
Update MSRV to 1.70

Doods0:
Debian stable [now supports](https://packages.debian.org/trixie/rust/rustc) rustc 1.85, so I bumped the MSRV to 1.70 to remove `once_cell`.

```
# Not available in stdlib until 1.70, but we support 1.63 to support Debian stable.
once_cell = { version = "1.19.0", default-features = false, features = ["std"] }
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         rust-version:
           - nightly
           - stable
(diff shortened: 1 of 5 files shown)


Doods0:
Apparently, the maintainer is already planning another PR, #327 , so I'll close this.

